PR #17, opened 2019-08-05
Two leftover StreamField migrations

KalobTaulien:
When setting up the project I realized there were a couple migrations that weren't run. Figured they could go into the repo. I'd also be happy to run `black` on these files.


kaedroho:
Thanks @KalobTaulien, I don't seem to get these changes when I run `makemigrations` on master, could you try this again?

KalobTaulien:
Definitely, I'll let you know what happens shortly.

KalobTaulien:
@kaedroho This is a good catch. I re-cloned the repo and when I ran `makemigrations` in a new running instance, no changes were detected.

PR #18, opened 2019-08-05
Version bump to 2.2

KalobTaulien:
Includes required migration files. Didn't upgrade Django yet. I figure we could do these upgrades incrementally in case there's an issue in the upgrade path that someone discovers.


kaedroho:
Thanks!

PR #19, opened 2019-08-14
Switch staging environment to Heroku

tomusher:
We've moved the staging environment for wagtail.io to Heroku - this MR removes references to the old Dokku environment and updates the `fabfile` to support deploying to Heroku staging.


kaedroho:
Looks good, thanks!

PR #23, opened 2019-11-09
update blog.scss - ensure menu hides on sm devices

lb-:
* Note: quick fix
* due to the long list of blog posts, -200% is not enough to 'hide' all the blog listing
* ideally the blog listing should be completely hidden on mobile, however, display: none causes some flickering of the menu on scrolling


tomdyson:
Thanks @lb- !

Issue #20, opened 2019-10-04
Black formatting?


KalobTaulien:
Not an issue, but more of a discussion point. 

What are everyones thoughts on using **Black** to format Python files? I'm working on a couple updates to this repo and thought to myself, "While I'm working on specific files, I could format them as a separate commit."

kaedroho:
Hey Kalob, I'm in favour of this!

thibaudcolas:
I think people at Torchbox are beyond debating this – we just use it on all (new) projects at this stage. @KalobTaulien if you’re keen to add this to wagtail.io I think it would be very welcome, but since that PR would be likely to cause lots of merge conflicts I would suggest you identify someone to review it beforehand to make sure this spends the least amount of time possible un-merged.

PR #29, opened 2020-01-20
Black formatting

KalobTaulien:
PR for Issue #20 

Seeing as there's only one open PR right now and it only have 2 slightly modified Python files, I thought this would be a good time to run `black`.

Ran `black wagtailio/ --exclude /migrations/` from the source directory. Didn't bother running black on the migration files though. Although we can `black` those files too if someone feels like it's a good idea (I can't imagine people are modifying the existing migration files, but I could be wrong!)


kaedroho:
Thanks!

PR #32, opened 2020-02-10
Add newsletter RSS feed

### LINKED ISSUE #10: RSS feed for this week in Wagtail

Stormheg:
This fixes #10 

Feedback appreciated

### DIFF
--- a/wagtailio/newsletter/feeds.py
+++ b/wagtailio/newsletter/feeds.py
@@ -1,0 +1,32 @@
+from datetime import datetime, time
+
+from django.contrib.syndication.views import Feed
+from django.http import Http404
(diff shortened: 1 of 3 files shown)


kaedroho:
Hey @Stormheg thanks for adding this!

I've not looked into it much yet, but we probably should link to this from somewhere. I think there might be a `<meta>` tag we can use to let browsers know about it

Stormheg:
Good thinking! Will look into that!

Stormheg:
@kaedroho I've added a `<link rel="alternate" ...>` tag for the newsletter feed in `base.html`. This is also how the existing blog feed makes it presence known.

PR #33, opened 2020-02-10
Fix migrate command on release

### LINKED ISSUE #28: Migrations not being run on deployment

Stormheg:
This fixes #28 

Migrations were not being run because Heroku apparently only accepts a single release command entry anymore. Not sure why but with this fix it should be working again!


kaedroho:
Confirmed fixed on staging now, thanks!

PR #41, opened 2020-03-31
Hide footer on blog pages

Stormheg:
This PR will remove the footer from blog pages as it currently overlaps with the navigation when scrolled all the way to down:
image


Stormheg:
closing, superseded by #44

PR #53, opened 2020-07-02
Add Wagtail Alt Generator

nimasmi:
@tomdyson already added the required credentials added as environment variables to the production server


tomdyson:
The environment variables exist, but the settings are missing.

PR #54, opened 2020-07-02
Add wagtail-content-import

nimasmi:
This uses a non-standard docx parser, configured in settings, and a
custom docx image converter. Both will likely be redundant once the
corresponding features are made prat of wagtail-content-import itself.

I have added the corresponding credentials as environment variables to the production server.


wagtailio/settings/base.py
```diff
@@ -415,3 +418,5 @@
             "ZONEID": env["FRONTEND_CACHE_CLOUDFLARE_ZONEID"],
         }
     }
+
+WAGTAILCONTENTIMPORT_DOCX_PARSER = "wagtailio.utils.parsers.DocxHTMLParser"
```
jacobtoppm, line 422: I can't actually see this file in the code - is it possible you accidentally didn't commit it?

nimasmi: Thanks. I've added it now.

PR #64, opened 2020-10-06
Import/Export Apps Setup

jacobtoppm:
- Add Wagtail Image Import
- Update to 2.10
- Set up airtable sync for FeatureDescription and BlogPage models


tomdyson (approved):
Thanks, @jacobtoppm ! I would have slightly preferred these to be two separate PRs, as a simpler guide for future devs adding these features, but that was my fault for not specifying it, and I'm happy to go ahead with this.

PR #71, opened 2020-10-13
Add the ability to publish or unpublish when updating from airtable

jacobtoppm:


zerolab:
@jacobtoppm is this still needed?

jacobtoppm:
@zerolab It looks like main still doesn't have the publish on import, but I think this was left because it was a little ambiguous whether this was a feature we actually wanted beyond needing it for a WNiW demo. That said, I do think it would be good to get staging and main more in sync.

If we still want it, happy to make another PR without the unrelated commits that seem to have been dragged in here.

zerolab:
Will close as we will no longer use this here.

Best to do the demos on demo sites

PR #83, opened 2021-01-29
Blog items overlapping blog content issue

BrianXu20:
Apply the following fix to prevent blog-items from showing up on mobile view.

https://github.com/wagtail/wagtail.io/issues/77#issuecomment-730525297

### DIFF
--- a/wagtailio/static/css/components/blog.scss
+++ b/wagtailio/static/css/components/blog.scss
@@ -162,21 +162,21 @@
     z-index: 4;
     opacity: 0;
     transition: transform 0.2s ease 0s, opacity 0.2s ease 0s;
-    pointer-events: none;
+    overflow: hidden;
 
     @include media-query(small) {
       transform: translate3d(0, 0%, 0);
       position: relative;
       background-color: transparent;
       opacity: 1;
-      pointer-events: auto;
+      overflow: auto;
     }
 
     .sidebar-open & {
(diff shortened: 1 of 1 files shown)


chris-lawton:
Correct me if wrong but I think this would output the following which wouldn't work as `sidebar-open` isn't a child of `blog-listing__link`
```
 .blog-listing__link .sidebar-open:not {
      pointer-events: none; 
}
```

A different approach could be to add an `overflow:hidden` to the `blog-listing__container` and then allow scroll again when the sidebar is open:
`blog.scss`:
```
  &__container {
    position: absolute;
    width: 100%;
    height: 100%;
    transform: translate3d(0, -300%, 0);
    background-color: $color--headings;
    z-index: 4;
    opacity: 0;
    transition: transform 0.2s ease 0s, opacity 0.2s ease 0s;
    overflow: hidden; // new

    @include media-query(small) {
      transform: translate3d(0, 0%, 0);
      position: relative;
      background-color: transparent;
      opacity: 1;
      overflow: auto; // new
    }

    .sidebar-open & {
      transform: translate3d(0, 0%, 0);
      opacity: 1;
      padding-top: 127px;
      overflow: auto; // new
    }
  }
```
What do you think?

chris-lawton:
Thanks @BrianXu20!

chris-lawton (approved):

PR #87, opened 2021-02-03
Upgrade Python to 3.8

tomdyson:
to be consistent with wagtail-kit:

https://git.torchbox.com/internal/wagtail-kit/-/blob/master/wagtailkit_repo_name/Dockerfile#L19

Untested!

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Build Python app.
-FROM python:3.6.5-stretch
+FROM python:3.8-buster
 
 WORKDIR /app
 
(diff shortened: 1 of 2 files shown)


zerolab (approved):

zerolab (changes requested):
Tested locally, with a Python 3.8 in the buster VM.
Needs `psycopg2-binary` to work

tm-kn:
@zerolab Why `psycopg2-binary`? What for?

zerolab:
@tm-kn no reason other than it being in my clipboard 🤦‍♂️

`psycopg2` is the recommended package for production use. Pushed to staging.

zerolab (approved):

zerolab:
fixed via #100

PR #93, opened 2021-03-26
Add A/B testing

KalobTaulien:
We want to be able to test wagtail-ab-testing a bit more, and maybe run some tests on the wagtail.io site for fun. So I added the wagtail a/b testing package to the site. Tested on a fresh install and it works out of the box.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -24,6 +24,7 @@ wagtail-content-import==0.4.2
 wagtail-image-import==0.1.0
 wagtail-transfer==0.5.1
 mammoth==1.4.10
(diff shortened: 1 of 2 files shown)


tomdyson (commented):
LGTM. But will it work behind Cloudflare? 

edit: looks like it'll need https://github.com/torchbox/wagtail-ab-testing/pull/23

KalobTaulien:
@tomdyson I merged that work you mentioned, thanks for remembering about that! I version bumped wagtail-ab-testing to 0.2.1 and updated this PR with wagtail-ab-testing==0.2.1.

PR #100, opened 2021-04-24
Update dependencies

zerolab:
This PR updates a number of dependencies, including sentry

Closes #92, #94, #96, #87

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,31 +1,33 @@
-Django==2.2.13
-dj-database-url==0.5.0
+Django==2.2.20
 wagtail==2.12.4
-psycopg2==2.7.4
-libsass==0.8.3
-Pillow==7.1.2
+dj-database-url==0.5.0
+psycopg2==2.8.6
+libsass==0.20.1
+Pillow==8.1.2
 embedly==0.5.0
-Markdown==2.6.11
-Pygments==2.2.0
+Markdown==3.3.4
+Pygments==2.7.4
 django-compressor==2.2
 django-libsass>=0.7
-raven==6.7.0
-whitenoise==3.3.1
+whitenoise==5.2.0
 brotlipy==0.7.0
 django-storages==1.6.6
 boto3==1.7.82
 django-redis==4.9.0
+sentry-sdk==1.0.0
+scout-apm==2.18.0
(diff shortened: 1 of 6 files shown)


requirements.txt
```diff
@@ -1,31 +1,34 @@
...
-raven==6.7.0
-whitenoise==3.3.1
+whitenoise==5.2.0
 brotlipy==0.7.0
 django-storages==1.6.6
 boto3==1.7.82
 django-redis==4.9.0
+sentry-sdk==1.0.0
+scout-apm==2.18.0
+fabric3==1.14.post1
```
Stormheg, line 19: You've added this to `requirements-dev.txt` as well. Did you mean to add it here in the base requirements?

wagtailio/settings/base.py
```diff
@@ -397,19 +392,44 @@
 }
 
 
-# Sentry
+# Sentry configuration.
+# See instructions on the intranet:
+# https://intranet.torchbox.com/delivering-projects/tech/starting-new-project/#sentry
```
Stormheg, line 397: This is an internal Torchbox link. Probably best to remove it.

wagtailio/settings/base.py
```diff
@@ -397,19 +392,44 @@
...
+    from sentry_sdk.utils import get_default_release
+
+    sentry_kwargs = {
+        "dsn": env["SENTRY_DSN"],
+        "integrations": [DjangoIntegration()],
+    }
+
+    # There's a chooser to toggle between environments at the top right corner on sentry.io
+    # Values are typically 'staging' or 'production' but can be set to anything else if needed.
+    # dokku config:set gosh SENTRY_ENVIRONMENT=staging
```
Stormheg, line 413: `gosh` is Torchbox client app I think? Probably better to remove this line as wagtail.io is not hosted using dokku.

Stormheg (approved):
Thanks @zerolab! Wagtail.io needs this kind of love :heart: 
I've added some small comments for things I think can be removed.

Tested this on my machine with the new vagrant box and staging data and everything appears to work correctly. I'm not able to verify the sentry configuration but I trust it to be correct.

requirements.txt
zerolab: nice catch. borked rebase

PR #102, opened 2021-05-28
Upgrading wagtail.io packages

KalobTaulien:
May need some testing with production data. But locally this spins up fine using vagrant and migrations get applied fine as well.


zerolab (commented):
@KalobTaulien thanks for this. Would be nice to bump Django to 3.2 as well

zerolab:
Closing in favour of #133

PR #111, opened 2021-07-16
Add ability to insert wagtailmedia videos into features

kaedroho:


wagtailio/features/models.py
```diff
@@ -43,6 +51,7 @@ def __str__(self):
         InlinePanel("bullets", label="Bullets"),
         ImageChooserPanel("screenshot"),
         FieldPanel("video_url"),
+        MediaChooserPanel('video'),
```
zerolab, line 54: I've opened an issue on wagtailmedia - https://github.com/torchbox/wagtailmedia/issues/133- to allow more targeted choosers

zerolab (approved):

wagtailio/features/models.py
kaedroho: That would be awesome, thanks!

PR #112, opened 2021-07-22
Add video option to BannerBlock

kaedroho:
This adds an option to insert a video on the homepage banner. Requested by Will


wagtailio/core/blocks.py
```diff
@@ -64,13 +66,31 @@ class Meta:
...
+    def clean(self, value):
+        errors = {}
+
+        if not value.get('image') and not value.get('video'):
+            error_message = 'You must select either an image or a video.'
+            errors['image'] = ErrorList([error_message])
+            errors['video'] = ErrorList([error_message])
+
+        if errors:
+            raise ValidationError('Validation error in StructBlock', params=errors)
```
zerolab, line 90: question: does "Validation error in StructBlock" get shown to the user?
if so, a more suggestive message would be more helpful. If not, do ignore me :)

wagtailio/core/blocks.py
```diff
@@ -64,13 +66,31 @@ class Meta:
         template = "core/blocks/page_or_external_link_block.html"
 
 
+class BannerVideoBlock(media_blocks.AbstractMediaChooserBlock):
+    pass
```
zerolab, line 70: suggestion:

```suggestion
    @cached_property
    def field(self):
        return forms.ModelChoiceField(
            queryset=self.target_model.objects.filter(type="video"),
            widget=self.widget,
            required=self._required,
            validators=self._validators,
            help_text=self._help_text,
        )
```

to limit just to videos

wagtailio/core/blocks.py
kaedroho: This doesn't get displayed anywhere, in fact, Wagtail uses this as the error message too: https://github.com/wagtail/wagtail/blob/main/wagtail/core/blocks/struct_block.py#L20

Note that ``StructBlockValidationError`` was introduced recenly and we can't use that on wagtail.io yet

wagtailio/core/blocks.py
kaedroho: Thanks! Manually applied to add imports

wagtailio/core/blocks.py
zerolab: thank you @kaedroho, should not have been lazy!

zerolab (approved):
:shipit:

PR #129, opened 2022-01-19
[Areweheadlessyet] Add areweheadlessyet app and initial HomePage model

Tijani-Dia:
This PR adds a `areweheadlessyet` app and basic HomePage model for it.


wagtailio/areweheadlessyet/apps.py
```diff
@@ -0,0 +1,6 @@
+from django.apps import AppConfig
+
+
+class AreweheadlessyetConfig(AppConfig):
```
thibaudcolas, line 4: Shouldn’t this be:

```suggestion
class AreWeHeadlessYetConfig(AppConfig):
```

to match what you’ve used in the models / follow Django conventions?

wagtailio/areweheadlessyet/models.py
```diff
@@ -0,0 +1,34 @@
...
+from wagtail.core.models import Page
+
+
+class AreWeHeadlessYetHomePage(Page):
+    THUMBS_UP = "\U0001f44D"
+    THUMBS_DOWN = "\U0001f44E"
+    ICON_CHOICES = [
+        (THUMBS_UP, THUMBS_UP),
+        (THUMBS_DOWN, THUMBS_DOWN),
+    ]
```
thibaudcolas, line 14: I think we should store icon choices with string identifiers (`thumbsup`, `thumbsdown`) rather than emoji codes, and leave it up to the front-end to render either vanilla emojis or image files.

wagtailio/areweheadlessyet/models.py
```diff
@@ -0,0 +1,34 @@
+from django.db import models
+
+from wagtail.admin.edit_handlers import FieldPanel, MultiFieldPanel
+from wagtail.core.fields import RichTextField
+from wagtail.core.models import Page
+
+
+class AreWeHeadlessYetHomePage(Page):
```
thibaudcolas, line 8: Should we reuse this project’s `SocialMediaMixin`? And potentially other mixins?

wagtailio/areweheadlessyet/models.py
Tijani-Dia: Alright, I've updated it. Is that what you meant?

wagtailio/areweheadlessyet/models.py
Tijani-Dia: I've added it along with the `CrossPageMixin`.

wagtailio/areweheadlessyet/models.py
thibaudcolas: Yep!

RealOrangeOne (approved):

PR #130, opened 2022-01-19
[Areweheadlessyet] Add base section/information block

Tijani-Dia:
This PR adds a generic SectionBlock to be used in the HomePage body.

It's based on #129. I will rebase when appropriate.


wagtailio/areweheadlessyet/models.py
```diff
@@ -0,0 +1,37 @@
...
+
+class AreWeHeadlessYetHomePage(Page):
+    THUMBS_UP = "\U0001f44D"
+    THUMBS_DOWN = "\U0001f44E"
+    ICON_CHOICES = [
+        (THUMBS_UP, THUMBS_UP),
+        (THUMBS_DOWN, THUMBS_DOWN),
+    ]
+    strapline_icon = models.CharField(
+        max_length=255,
```
RealOrangeOne, line 17:
```suggestion
        max_length=1,
```

Although, with unicode codepoints, 1 may not actually be correct. Either way 255 is probably way too high

RealOrangeOne (approved):

RealOrangeOne:
> It's based on #129 therefore only the second commit needs reviewing. I will rebase when appropriate.
I would recommend using a feature branch, and merging things into that. Makes the rebasing hassle a bit easier, and can make reviewing much easier, too!

Tijani-Dia:
> I would recommend using a feature branch, and merging things into that. Makes the rebasing hassle a bit easier, and can make reviewing much easier, too!

I had the same thought. Can you create that feature branch please?

wagtailio/areweheadlessyet/models.py
```diff
@@ -0,0 +1,42 @@
...
+        ),
+        StreamFieldPanel("body"),
+    ]
+
+    promote_panels = (
+        Page.promote_panels + SocialMediaMixin.panels + CrossPageMixin.panels
+    )
+
+    max_count = 1
+    parent_page_types = ["wagtailcore.Page"]
```
zerolab, line 41: question: what is the purpose for this? could use a comment for future devs

Tijani-Dia: The `parent_page_types` line is added so that we can only create the AreWeHeadlessYetHomepage at the root level (another site instance).
The `max_count` line is here for ensuring that we are can add one and only one AreWeHeadlessYetHomePage.

Does it make sense?

zerolab: Hey @Tijani-Dia 👍🏼 all makes sense to me. As I mentioned, worth adding a comment to that effect

Tijani-Dia: Added!

PR #131, opened 2022-01-20
[AreWeHeadlessYet] News section

Tijani-Dia:
This PR adds a news section to AreWeHeadlessYet HomePage model.

We can pick 3 blog pages from the wagtail.org blog to display as news in areweheadlessyet site.


wagtailio/areweheadlessyet/models.py
```diff
@@ -1,12 +1,36 @@
...
+    page = ParentalKey(
+        "areweheadlessyet.AreWeHeadlessYetHomePage",
+        on_delete=models.CASCADE,
+        related_name="news",
+    )
+    blog_post = models.ForeignKey(
+        "blog.BlogPage",
+        null=True,
+        blank=True,
+        on_delete=models.SET_NULL,
```
RealOrangeOne, line 27: Suggestion: I think in this case we actually want `models.CASCADE`. Otherwise we're left with a `AreWeHeadlessYetNews` without a related page, which could cause a lot of issues. This way, the `AreWeHeadlessYetNews` is just deleted too, and it just disappears from the homepage, which I is probably what a user would expect anyway

Tijani-Dia: Alright! I've updated it.

RealOrangeOne (approved):

PR #135, opened 2022-01-24
[AreWeHeadlessYet] Add news section in HomePage's body

Tijani-Dia:
We added a news section in #131 however, after some testing, we'd prefer to have that section inside the page's body.


wagtailio/areweheadlessyet/migrations/0001_initial.py
```diff
@@ -1,89 +1,36 @@
-# Generated by Django 3.2.8 on 2022-01-19 12:58
```
RealOrangeOne, line 1: Note: Editing migrations is generally a bad idea, as this won't be run during deployment. If we want this to run, you will need to roll back migrations after deployment, before which the migrations won't actually have run.

Tijani-Dia: Do you suggest that I keep the previous migrations and just add another one?

RealOrangeOne: Given we're so early in the project, let's keep the history clean and keep it as you have it now (ie 1 new commit and remove the old one). It's just something to be aware of when deploying.

PR #136, opened 2022-01-25
[AreWeHeadlessYet] Issues section

Tijani-Dia:
This PR adds an issue section in the HomePage body.

A management command will be scheduled to run once a day to fetch issues from Github's API tagged #headless in the Wagtail repository.


wagtailio/areweheadlessyet/management/commands/fetch_headless_issues.py
```diff
@@ -0,0 +1,64 @@
...
+        )
+
+        if not response.ok:
+            message = response.json()["message"]
+            self.stdout.write(
+                self.style.ERROR(
+                    f"Fetching from Github's API failed for the following reason: {message}"
+                )
+            )
+            return
```
RealOrangeOne, line 23:
```suggestion
        response.raise_for_status()
```

Suggestion: We can simplify this down quite a lot, and just raise an exception instead. This has the added benefit of reporting back into Sentry in the event of an issue, rather than just the logs. Along with said exception, it also contains a bunch of context about the response and error, which would be useful for debugging.

wagtailio/areweheadlessyet/management/commands/fetch_headless_issues.py
```diff
@@ -0,0 +1,64 @@
...
+            WagtailHeadlessIssue.objects.values_list("number", flat=True)
+        )
+        to_delete, to_create = ([], [])
+        for issue in issues:
+            if (issue_number := issue["number"]) in issues_in_db:
+                # Delete issue when it's closed.
+                if issue["state"] == "closed":
+                    to_delete.append(issue_number)
+            else:
+                to_create.append(
```
RealOrangeOne, line 36: Suggestion: I notice this doesn't handle title changes for issues. If that's something we want, it might be better to use `create_or_update` here. That will slow things down as it'll do `n` queries, but with the numbers we're dealing with, it shouldn't make a huge difference.

wagtailio/areweheadlessyet/models.py
```diff
@@ -2,11 +2,24 @@
...
 
 from wagtailio.areweheadlessyet.blocks import HomePageBlock
 from wagtailio.utils.models import CrossPageMixin, SocialMediaMixin
 
 
+@register_snippet
+class WagtailHeadlessIssue(models.Model):
+    number = models.PositiveIntegerField(primary_key=True)
+    title = models.CharField(max_length=255)
+    url = models.URLField()
```
RealOrangeOne, line 15: Question: The URL can always be derived from the issue number, is there a need to store it specifically?

eg:
```python
@property
def url(self):
    return f"https://github.com/wagtail/wagtail/issues/{self.number}/"
```

wagtailio/areweheadlessyet/management/commands/fetch_headless_issues.py
Tijani-Dia: Alright, it's added.

wagtailio/areweheadlessyet/management/commands/fetch_headless_issues.py
Tijani-Dia: Alright.

wagtailio/areweheadlessyet/models.py
Tijani-Dia: I agree. I have changed it.

RealOrangeOne (approved):

PR #142, opened 2022-02-10
[AreWeHeadlessYet] Use a more user-friendly date format

Tijani-Dia:
Use a more user-friendly date format for blog posts API representation

### DIFF
--- a/wagtailio/areweheadlessyet/blocks.py
+++ b/wagtailio/areweheadlessyet/blocks.py
@@ -29,7 +29,7 @@ class BlogPostChooserBlock(PageChooserBlock):
         if value is None:
             return None
         return {
-            "date": value.date,
+            "date": value.date.strftime("%d %b %Y"),
             "title": value.title,
             "url": value.full_url,
             "introduction": value.introduction,


wagtailio/areweheadlessyet/blocks.py
```diff
@@ -29,7 +29,7 @@ def get_api_representation(self, value, context=None):
         if value is None:
             return None
         return {
-            "date": value.date,
+            "date": value.date.strftime("%d %b %Y"),
```
RealOrangeOne, line 32: I'd actually suggest not doing this. "formatting" like this should be left down to the frontend / API consumer, rather than being mandated by the API. ISO8601 is easy to parse, and supported by basically everything. Formatting in the frontend allows more customization, and makes it possible to change the representation in different parts of the application. It would also allow using [locale-specific formats](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleDateString).

Tijani-Dia: That makes sense. Thanks!

PR #147, opened 2022-03-02
Add sitewide announcement block

zerolab:
This PR adds a new `sitewide_alerts` app which registers a setting to allow enabling sitewide alerts.

It uses JS to fetch it so as to bypass FE cache. The enpoint is cached and will purge the FE cache on save.

This is how it looks on the homepage when enabled. The styling could probably use some work

Screenshot 2022-03-02 at 21 05 32

Screenshot 2022-03-02 at 22 34 54

with custom colours:

bg: `#333333`, fg: `#ffffff`
Screenshot 2022-03-02 at 21 47 43

Screenshot 2022-03-02 at 22 35 31

bg: `#0057B8`, fg: `#FFD700`
Screenshot 2022-03-02 at 21 48 33
Screenshot 2022-03-02 at 22 35 17


Scotchester:
Couple immediate thoughts:
1. Wouldn't it be more customary to place this sort of banner above everything, including the header?
2. It'd be nice to style this with the colors of the people we are immediately wanting to use it to support. Maybe add fields for background color and foreground color? Or just a field to enter some arbitrary CSS?

zerolab:
@Scotchester 

1. the sticky menu is a massive !@#$ pain in the arse. Went with the best solution I had
2. a good shout. we'd need to handle text colour as well as bg.

KalobTaulien:
I see why the banner is under the nav. This is the fastest implementation without getting into a rabbit hole of sticky navbars, especially when moving up and down the page and _maybe_ having the banner show up with the nav (or not). We can always adjust that later down the road if needed, too. 

I agree with the background and text colors - we might want to make these adjustable so we can re-use the banner for other (or more specific) purposes in the future. 

I'm not opposed to getting this accepted ASAP to help get our support and messaging out faster though.

thibaudcolas:
Same! I’ll get this running locally too to test.

Scotchester:
Ah yeah, I forgot the nav was sticky when commenting. Given that, I'm fine with the placement.

zerolab:
Updated with settings for bg/text colours and provided screenshots (in the first comment)

Scotchester:
Is there supposed to be bold text in those screenshots?

zerolab:
gah, I forgot to actually enbolden the word. updated screenshots with bold and italic

Scotchester (approved):
This front-end rendering approach is new to me, but I don't have any experience with the codebase of this particular site, so it may make complete sense to others :)

Approving from the standpoint that it looks to provide the minimum functionality we need to get a message out there quickly.

wagtailio/static/js/main.js
```diff
@@ -160,4 +160,23 @@ $(function() {
...
+          'top': $('.header')[0].getBoundingClientRect().bottom + 'px'
+        };
+        if (data.bg_colour) {
+          styles['background-color'] = '#' + data.bg_colour;
+        }
+        if (data.text_colour) {
+          styles['color'] = '#' + data.text_colour;
+        }
+        $alert_container.css(styles).toggleClass('alert--default', !(data.bg_colour || data.text_colour));
+        $alert_container.prop('aria-expanded', true).addClass('alert--active');
```
thibaudcolas, line 178:
```suggestion
        $alert_container.addClass('alert--active');
```

zerolab:
In some ways it probably is overkill, but purging one url from cache vs all of them felt better. At the same time, there's an additional http request so ⚖️

wagtailio/static/js/main.js
```diff
@@ -160,4 +160,23 @@ $(function() {
     })
   }
 
+  var $alert_container = $("[data-alert]");
+  if ($alert_container.data("alert")) {
+    $.getJSON($alert_container.data("alert"), function(data) {
+      if (data.text) {
+        $alert_container.html(data.text);
+        var styles = {
+          'top': $('.header')[0].getBoundingClientRect().bottom + 'px'
```
thibaudcolas, line 169: For me this makes the alert display partly above the next element. Not an issue with the homepage but might be problematic elsewhere.

thibaudcolas (approved):

thibaudcolas:
@zerolab for your consideration: #148. I moved the banner above the header too (can be moved back down easily if we prefer).

PR #148, opened 2022-03-02
Remove the need for relative positioning causing layout issues

thibaudcolas:
#147 with extra front-end tweaks. The alert itself looks the same, I just fixed the issues I reported in #147, fixed the animation, and moved it to be above the header.

If we want to move it below the header again, it’s just a matter of moving the element back to where it was before.

sitewide


zerolab:
🌟🌟🌟

PR #150, opened 2022-03-03
Modify the API to serve content from AreWeHeadlessYet site

Tijani-Dia:
**Context**

We want to query data from the AreWeHeadlessYet site using the API. However, we can't use the `areweheadlessyet.wagtail.org` domain for the  AreWeHeadlessYet site/backend because it's already used for the Next.js frontend hosted in Vercel.

**Solution**

Use the `wagtail.org` domain to serve content from the AreWeHeadlessYet site/backend - that is the goal of this PR.

### DIFF
--- a/wagtailio/api.py
+++ b/wagtailio/api.py
@@ -11,14 +11,13 @@ class AreWeHeadlessYetPagesAPIViewSet(PagesAPIViewSet):
     def get_base_queryset(self):
         """Returns a queryset containing only pages from the AreWeHeadLessYet site."""
 
-        areweheadlessyet_queryset = AreWeHeadlessYetHomePage.objects.all()
-        if not areweheadlessyet_queryset:
-            return areweheadlessyet_queryset
-
-        areweheadlessyet_root_page = areweheadlessyet_queryset[0]
-        return Page.objects.live().descendant_of(
-            areweheadlessyet_root_page, inclusive=True
-        )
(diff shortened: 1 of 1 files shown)


wagtailio/api.py
```diff
@@ -1,8 +1,24 @@
...
+
+class AreWeHeadlessYetPagesAPIViewSet(PagesAPIViewSet):
+    def get_base_queryset(self):
+        """Returns a queryset containing only pages from the AreWeHeadLessYet site."""
+
+        areweheadlessyet_queryset = AreWeHeadlessYetHomePage.objects.all()
+        if not areweheadlessyet_queryset:
+            return areweheadlessyet_queryset
+
+        areweheadlessyet_root_page = areweheadlessyet_queryset[0]
```
kaedroho, line 18: You could use ``.first()`` here which will return ``None`` if the queryset is empty so you wouldn't need to perform the query above to check if there are any pages in the queryset first:

```python
areweheadlessyet_root_page = areweheadlessyet_queryset.first()

if areweheadlessyet_root_page:
    return Page.objects.live().descendant_of(
        areweheadlessyet_root_page, inclusive=True
    )
else:
    return Page.objects.none()
```

kaedroho (changes requested):
This looks good to me, thanks!

Just one quite minor tweak

Tijani-Dia:
Thanks, @kaedroho.

I addressed your comment.

kaedroho (approved):

PR #157, opened 2022-04-26
Trigger deployment on Vercel when a page in areweheadlessyet is (un)published

Tijani-Dia:
The goal of this PR is to trigger a deployment in Vercel when a page is published/unpublished in the areweheadlessyet site.

This is heavily inspired by https://github.com/tomdyson/wagtail-netlify/blob/master/wagtailnetlify/models.py.

Do we want to track the status of deployments?


wagtailio/areweheadlessyet/signal_handlers.py
```diff
@@ -0,0 +1,39 @@
...
+        t.start()
+
+    return decorator
+
+
+@postpone
+def deploy(sender, **kwargs):
+    """Triggers a build on Vercel."""
+
+    response = requests.post(settings.VERCEL_DEPLOY_HOOK_URL)
```
kaedroho, line 27: We should add a timeout here, in case there is a network issue (requests will wait forever by default)

wagtailio/areweheadlessyet/signal_handlers.py
```diff
@@ -0,0 +1,39 @@
...
+from .models import AreWeHeadlessYetHomePage, AreWeHeadlessYetTopicPage
+
+
+def postpone(function):
+    """
+    Cheap aysnc, see https://stackoverflow.com/a/28913218
+    """
+
+    def decorator(*args, **kwargs):
+        t = Thread(target=function, args=args, kwargs=kwargs)
```
kaedroho, line 16: If someone publishes/unpublishes a lot of pages at once, this could create a lot of threads. Could we use a thread-pool here to limit the number of threads that can be running?

kaedroho (changes requested):

wagtailio/settings/production.py
```diff
@@ -28,6 +28,10 @@
 
 FB_APP_ID = 710333029076622
 
+VERCEL_DEPLOY_HOOK_URL = env.get("VERCEL_DEPLOY_HOOK_URL", None)
+VERCEL_DEPLOY_REQUEST_TIMEOUT = env.get("VERCEL_DEPLOY_REQUEST_TIMEOUT", 0.001)
```
kaedroho, line 32: This is an interesting idea!

If we don't wait longer than 1ms for a response, we probably don't need to run it in a thread! But should we catch the timeout error though?

Tijani-Dia: Do I increase the timeout or remove the threads?

kaedroho: It might be fine to leave it something really small (like 1 second) as long as we catch the timeout error. We don't actually do anything with the response so I don't think we need to wait for it.

kaedroho (approved):

PR #158, opened 2022-04-27
Align text to be centered in sitewide setting

KalobTaulien:
Simply centers the text in the sitewide setting for a small aesthetic update. 

Screenshot below:
image


thibaudcolas:
I expect this is no longer relevant with the current redesign so will close this now.

PR #121, opened 2021-11-02
Configuration for docker-compose and gitpod

kaedroho:
This PR adds configuration for docker-compose and gitpod.

Looking to get some initial feedback, I'll update the readme and remove Vagrant if everyone is happy with this config.

PR #161, opened 2022-05-05
Remove vagrant setup for development

nickmoreton:
### This removes Vagrant developer setup along with associated files.

In its place is a Docker for development setup.

Fabric is used to start up the development environment.

Docker can be used locally as well as in a Gitpod development environment.


docker-compose.yml
```diff
@@ -1,47 +1,65 @@
...
+
     depends_on:
-      db:
-        condition: service_healthy
+      - db
+      - redis
 
   db:
-    image: postgres:13
+    image: postgres:12.3
```
zerolab, line 44: nitpick: this sort of blanket change is... not ok.

e.g. staging is on 13.6 so 13 is much better than the downgrade

Dockerfile
```diff
@@ -1,72 +1,109 @@
...
+ARG POETRY_INSTALL_ARGS="--no-dev"
 
-# Install operating system dependencies.
-RUN apt-get update -y && \
-    apt-get install -y apt-transport-https rsync libmagickwand-dev unzip postgresql-client-13 && \
-    rm -rf /var/lib/apt/lists/*
+# IMPORTANT: Remember to review both of these when upgrading
+ARG POETRY_VERSION=1.1.13
+# To get this value locally:
+# $ wget https://raw.githubusercontent.com/python-poetry/poetry/1.1.8/get-poetry.py
```
zerolab, line 13:
```suggestion
# $ wget https://raw.githubusercontent.com/python-poetry/poetry/1.1.13/get-poetry.py
```
?

kaedroho:
I'm quite happy with removing Vagrant, but are the changes to Docker necessary? A lot of those configs were written a couple of months ago.

What problems are you solving by rewriting them?

Makefile
```diff
@@ -1,36 +0,0 @@
-.PHONY: setup rebuild start superuser migrations migrate pull-production-data pull-production-media pull-staging-data pull-staging-media
```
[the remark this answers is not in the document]
kaedroho: The setup command is quite important here as that sets the UID of the app container properly. This is necessary to prevent any files that Django writes from being owned by root.

docker-compose.yml
nickmoreton: Changed version to 13

pyproject.toml
```diff
@@ -0,0 +1,52 @@
+[tool.poetry]
+name = "wagtailio"
+version = "0.1.0"
+description = ""
+authors = ["Your Name <you@example.com>"]
```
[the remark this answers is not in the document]
nickmoreton: Added author

Dockerfile
nickmoreton: Corrected comment

Makefile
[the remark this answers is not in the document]
nickmoreton: I have restored the Makefile

docker-compose.yml
```diff
@@ -1,47 +1,65 @@
...
-      POSTGRES_PASSWORD: wagtailio
+      POSTGRES_PASSWORD: wagtailio # pragma: allowlist secret
       PGDATABASE: wagtailio
       PGUSER: wagtailio
-      PGPASSWORD: wagtailio
+      PGPASSWORD: wagtailio # pragma: allowlist secret
       POSTGRES_HOST_AUTH_METHOD: trust
     logging:
       driver: none
-    healthcheck:
```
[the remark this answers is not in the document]
nickmoreton: I've added the health check back in.

nickmoreton:
> I'm quite happy with removing Vagrant, but are the changes to Docker necessary? A lot of those configs were written only a couple of months ago. See #121

Thanks for the reviews both.

I can see what you mean @kaedroho is the first question here `should we add poetry` if so then the Dockerfile does need updating. If we don't add poetry then there's probably no need to do any of this as you updated it recently. Leaving Vagrant in place might also be a good idea for people who want to use that.

It's not a problem to not include this work at all if you prefer, I can make a note on the list.

Dockerfile
```diff
@@ -1,72 +1,109 @@
-# Build Python app.
-FROM python:3.8-bullseye AS backend
+# We use Debian images because they are considered more stable than the alpine
+# ones becase they use a different C compiler. Debian images also come with
+# all useful packages required for image manipulation out of the box. They
+# however weight a lot, approx. up to 1.5GiB per built image.
+FROM python:3.8 as production
```
kaedroho, line 5: Image should be ``python:3.8-bullseye``. see: https://github.com/wagtail/wagtail.org/pull/121#discussion_r788967227

Dockerfile
```diff
@@ -1,72 +1,109 @@
...
+ARG POETRY_INSTALLER_SHA=eedf0fe5a31e5bb899efa581cbe4df59af02ea5f
 
-WORKDIR /app
-EXPOSE 8000
+# Install dependencies in a virtualenv
+ENV VIRTUAL_ENV=/venv
 
-# Create a non-root application user.
-ARG UID=1000
-RUN useradd wagtailio -u $UID -m
```
kaedroho, line 21: Please could we keep the ``UID`` argument? This makes the development experience a bit nicer as any files that are created by Django would be given the same UID as the host user so there won't be any permissions issues.

nickmoreton:
Morning @kaedroho @zerolab 

I pretty sure the comments have been addressed here now. 

Do you have any further thoughts about this work? Thanks

zerolab:
@nickmoreton @kaedroho fwiw, we are moving to poetry due to many lose requirements.
This is in #265, which upgrades the build to Wagtail 4.

Will do separate PRs for the vagrant removal, and perhaps vscode devcontainer

PR #162, opened 2022-05-11
Move deploy signal handler outside function to avoid it being garbage collected

Tijani-Dia:
We were having an issue with the automatic deploy hook as it seemed that the signal listeners 'weren't listening'.

This happened because the `deploy` function was defined inside another function. Thus, no other part of the program references it once that function is run. However, signal listeners are weakly referenced hence the 'bug'.

I changed the `deploy` function to be a global one.

### DIFF
--- a/wagtailio/areweheadlessyet/thread_pool.py
+++ b/wagtailio/areweheadlessyet/thread_pool.py
@@ -26,3 +26,16 @@ def run_in_thread_pool(function):
         thread_pool.send((function, args, kwargs))
 
     return decorator
+
+
+@run_in_thread_pool
+def deploy(sender, **kwargs):
+    """Triggers a build on Vercel."""
+
+    try:
+        requests.post(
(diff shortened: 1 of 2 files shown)


wagtailio/areweheadlessyet/thread_pool.py
```diff
@@ -26,3 +26,17 @@ def decorator(*args, **kwargs):
...
+    """Triggers a build on Vercel."""
+
+    try:
+        requests.post(
+            settings.VERCEL_DEPLOY_HOOK_URL, 
+            timeout=settings.VERCEL_DEPLOY_REQUEST_TIMEOUT
+        )
+    except requests.exceptions.Timeout:
+        pass  # Ignore this error
+
```
kaedroho, line 42: Looks like git is complaining about no newline at the end of this file

kaedroho (approved):
This looks good to me thanks!

PR #163, opened 2022-05-25
Import requests and log deploy status

Tijani-Dia:
I forgot to update the imports after moving the `deploy` function in `thread_pool.py`.
I also added some logging lines.

### DIFF
--- a/wagtailio/areweheadlessyet/thread_pool.py
+++ b/wagtailio/areweheadlessyet/thread_pool.py
@@ -1,6 +1,10 @@
+import logging
 from concurrent.futures import ThreadPoolExecutor
 
+import requests
 from django.conf import settings
+
+logger = logging.getLogger("wagtailio")
 
 
 def run_thread_pool():
@@ -33,9 +37,17 @@ def deploy(sender, **kwargs):
     """Triggers a build on Vercel."""
 
     try:
-        requests.post(
+        logger.info("Triggering build on Vercel.")
+        response = requests.post(
             settings.VERCEL_DEPLOY_HOOK_URL,
             timeout=settings.VERCEL_DEPLOY_REQUEST_TIMEOUT,
         )
+        response.raise_for_status()
+        logger.info("Build triggered on Vercel.")
+
     except requests.exceptions.Timeout:
-        pass  # Ignore this error
+        logger.warning("The request to trigger a new build on Vercel has timed out.")
+        return  # Ignore this error
+
+    except Exception as e:
+        logger.warning(f"The request to trigger a new build on Vercel has failed: {e}.")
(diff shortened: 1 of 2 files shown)


wagtailio/areweheadlessyet/thread_pool.py
```diff
@@ -33,9 +37,17 @@ def deploy(sender, **kwargs):
...
+        response.raise_for_status()
+        logger.info("Build triggered on Vercel.")
+
     except requests.exceptions.Timeout:
-        pass  # Ignore this error
+        logger.warning("The request to trigger a new build on Vercel has timed out.")
+        return  # Ignore this error
+
+    except Exception as e:
+        logger.warning(f"The request to trigger a new build on Vercel has failed: {e}.")
```
zerolab, line 53: suggestion: use `logger.exception()` so the stack trace is captured

wagtailio/areweheadlessyet/thread_pool.py
```diff
@@ -33,9 +37,17 @@ def deploy(sender, **kwargs):
...
         )
+        response.raise_for_status()
+        logger.info("Build triggered on Vercel.")
+
     except requests.exceptions.Timeout:
-        pass  # Ignore this error
+        logger.warning("The request to trigger a new build on Vercel has timed out.")
+        return  # Ignore this error
+
+    except Exception:
```
kaedroho, line 52: Nitpick: Whenever writing ``except`` clauses that catch all exceptions, I'd recommend always adding a separate ``except`` clause to make sure ``KeyboardInterrupt`` and ``SystemExit`` don't get caught.

So just above this line, add:

```python
except (KeyboardInterrupt, SystemExit):
    raise
```

It's unlikely to be needed here, but I've found this to be a good practise to always add as sometimes it can be needed when you don't expect it!

kaedroho (approved):
Just a minor nitpick, but this looks good to me. Thanks!

wagtailio/areweheadlessyet/thread_pool.py
Tijani-Dia: Thanks!

wagtailio/areweheadlessyet/thread_pool.py
```diff
@@ -33,9 +37,20 @@ def deploy(sender, **kwargs):
     """Triggers a build on Vercel."""
 
     try:
-        requests.post(
+        logger.info("Triggering build on Vercel.")
+        response = requests.post(
             settings.VERCEL_DEPLOY_HOOK_URL,
             timeout=settings.VERCEL_DEPLOY_REQUEST_TIMEOUT,
```
zerolab, line 43: needs casting to int

PR #165, opened 2022-06-01
Expand rich text html for rich text fields

Tijani-Dia:
Someone noticed that links in RichtextField weren't processed. This PR aims to fix that.

### DIFF
--- a/wagtailio/areweheadlessyet/blocks.py
+++ b/wagtailio/areweheadlessyet/blocks.py
@@ -6,6 +6,7 @@ from wagtail.core.blocks import (
     StructBlock,
     URLBlock,
 )
+from wagtail.core.rich_text import expand_db_html
 from wagtail.snippets.blocks import SnippetChooserBlock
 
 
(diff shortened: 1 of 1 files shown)


wagtailio/areweheadlessyet/blocks.py
```diff
@@ -14,8 +15,13 @@ class LinkBlock(StructBlock):
     link_text = CharBlock(required=True)
 
 
+class _RichTextBlock(RichTextBlock):
```
zerolab, line 18: nitpick: 
```suggestion
class RichTextBlockWithExpandedApiRepresentation(RichTextBlock):
```

better to be a bit more verbose for the future.

Tijani-Dia: I was exactly going to do something like that but I was hesitating. Thanks, I won't hesitate to be verbose in the future.

Tijani-Dia: Thanks! Would you mind approving the PR if it's fine for you please?

zerolab (approved):

wagtailio/areweheadlessyet/blocks.py
zerolab: absolutely. nice one, @Tijani-Dia

PR #172, opened 2022-07-08
Adjusted display settings to prevent image overflow in textimage blocks

vossisboss:
Images were overflowing from the textimage blocks and were displaying poorly in mobile. I changed the overflow settings and added some media queries to make these blocks display better. Please let me know if you have any suggested changes.

### DIFF
--- a/wagtailio/static/css/components/text-image.scss
+++ b/wagtailio/static/css/components/text-image.scss
@@ -14,15 +14,27 @@
   .container {
     max-width: 800px;
     margin: 0 auto;
+    overflow: hidden;
 
     .text {
       max-width: 50%;
+
+      @include media-query(mobileOnly) {
+        max-width: 100%;
+        padding-top: 50px;
+      }
     }
 
     .image {
       float: right;
       max-width: 50%;
       padding-left: $gutter;
+
+      @include media-query(mobileOnly){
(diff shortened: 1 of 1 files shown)


zerolab:
@vossisboss could you add a representative link/screenshot?

vossisboss:
@zerolab Yup. These are some images from a draft of a new page I'm building. Here is how the images displayed in the block prior to this adjustment. Note how the image bleeds out into the text.
Screen Shot 2022-07-10 at 3 14 31 PM
And note how the image becomes so very tiny in a mobile-size display:
Screen Shot 2022-07-10 at 3 14 45 PM
Here is how the same content looks after the adjustment on a larger screen:
Screen Shot 2022-07-10 at 3 13 28 PM
And on a mobile screen, the image and the text will stack now:
Screen Shot 2022-07-10 at 3 13 45 PM

The most prominent use of this block in existing pages appears to be in the About page and it appears people corrected for the image bleed by adding additional text to expand the grey background. I would prefer not to add unnecessary text to avoid an image bleed.

thibaudcolas:
@vossisboss is this still relevant with our current launch of the updated site?

vossisboss:
This PR is no longer relevant given the move to the new design. Closing it out.

PR #179, opened 2022-07-25
Add stylelint config

chris-lawton:
- Adds rules from https://github.com/wagtail/wagtail/blob/main/.stylelintrc.js
- Ignore vendor dir
- Lint CSS

### DIFF
--- a/.stylelintrc.js
+++ b/.stylelintrc.js
@@ -1,3 +1,64 @@
 module.exports = {
   extends: '@wagtail/stylelint-config-wagtail',
+  rules: {
+    'scss/at-rule-no-unknown': [
+      true,
+      {
+        ignoreAtRules: [
+          'tailwind',
+          'apply',
+          'variants',
+          'responsive',
+          'screen',
+          'layer',
+        ],
+      },
+    ],
+    'no-invalid-position-at-import-rule': [
+      true,
+      {
+        ignoreAtRules: ['tailwind', 'use'],
+      },
+    ],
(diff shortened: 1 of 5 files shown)


.stylelintrc.js
```diff
@@ -1,3 +1,64 @@
 module.exports = {
   extends: '@wagtail/stylelint-config-wagtail',
+  rules: {
+    'scss/at-rule-no-unknown': [
+      true,
+      {
+        ignoreAtRules: [
+          'tailwind',
```
chris-lawton, line 8: Wonder if we should remove the tailwind references here as we (currently) don't use it.

William-Blackie: Yeah, we don't use it so no reason to include it.

.stylelintrc.js
```diff
@@ -1,3 +1,64 @@
...
+      },
+    ],
+    'no-invalid-position-at-import-rule': [
+      true,
+      {
+        ignoreAtRules: ['tailwind', 'use'],
+      },
+    ],
+    // Would be valuable for strict BEM components but is too hard to enforce with legacy code.
+    'no-descending-specificity': null,
```
William-Blackie, line 24: Note: As we're retooling, we could enforce it now?

William-Blackie (commented):
Hey, just a couple of tweaks I think @chris-lawton. Nice work.

.stylelintrc.js
chris-lawton: Sounds good, i'll remove it

PR #180, opened 2022-07-26
Remove code-highlighting css link from Blog

chris-lawton:
Removes references to `vendor/code-highlighting.css` as the design has updated for how these will look

### DIFF
--- a/wagtailio/blog/templates/blog/blog_index_page.html
+++ b/wagtailio/blog/templates/blog/blog_index_page.html
@@ -3,7 +3,6 @@
 {% load static wagtailcore_tags wagtailadmin_tags wagtailimages_tags wagtailio_utils %}
 
 {% block body_class %}template-blog{% endblock %}
-{% block extra_css %}<link rel="stylesheet" href="{% static 'css/vendor/code-highlighting.css' %}" />{% endblock %}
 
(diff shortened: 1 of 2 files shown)


wagtailio/blog/templates/blog/blog_index_page.html
```diff
@@ -3,7 +3,6 @@
 {% load static wagtailcore_tags wagtailadmin_tags wagtailimages_tags wagtailio_utils %}
 
 {% block body_class %}template-blog{% endblock %}
-{% block extra_css %}<link rel="stylesheet" href="{% static 'css/vendor/code-highlighting.css' %}" />{% endblock %}
```
zerolab, line 6: My preferred solution would be to bundle up code-highlighting.css so it can be used on the blog.

chris-lawton: Can do. I'd assumed that the new design would bring with it new code block styles but perhaps it's better to re-use this.

chris-lawton: I've added them to vendor which will be bundled with the main css file

zerolab (approved):
works for me =)

PR #181, opened 2022-07-26
Add new iconography

William-Blackie:
New icons available [here](http://0.0.0.0:8000/pattern-library/pattern/patterns/styleguide/icons.html)

Please note that this may not be an exhaustive list, just the vast majority. 

There's a small change in directory structure to add a display-only files directory.


William-Blackie:
Looks great @chris-lawton!

PR #185, opened 2022-07-26
Backend | New Streamfield blocks

engineervix:
This PR introduces new StreamField Blocks as follows:

- [x] CardsBlock
- [x] LogoCardsBlock
- [x] ComparisonTableBlock
- [x] HeadlineBlock
- [x] HighlightBlock
- [x] IconBulletsBlock
- [x] StandaloneQuoteBlock
- [x] MultipleQuotesBlock
- [x] RichTextBlock
- [x] StandAloneCTABlock
- [x] TeaserBlock
- [x] TextAndMediaBlock

The following blocks will be in a separate PR which will cover Snippets

- SignupFormBlock
- GetStartedChooserBlock

PR #188, opened 2022-07-28
Icon bullet

chris-lawton:
Adds icon bullet component viewable [locally in the pattern library](http://localhost:8000/pattern-library/pattern/patterns/components/streamfields/icon-bullet/icon-bullet.html).

Also:
- Adds base icon colours 
- Renames `icons.scss` to `_icons.scss`


wagtailio/project_styleguide/templates/patterns/components/streamfields/icon-bullet/icon-bullet.html
```diff
@@ -0,0 +1,18 @@
+{% load wagtailcore_tags %}
+<div class="icon-bullet grid">
```
zerolab, line 2: considering the block is a list if "icon bullet" items, should this not be `<li>`?

chris-lawton: It can be, we'd need to make the templates work for that. Assuming that this is the template that outputs an individual icon bullet streamfield `<li>` - we'd need a parent template to output the `ul` to wrap the children. I can do this in the pattern library but wanted to check that's ok to hook up on the BE? I wouldn't want the streamfield to reference this template directly so the `ul` would be missed off.

William-Blackie: I think the limitation of wagtail builder of not allowing nesting made this a problem. I agree with @zerolab, it's a nested data structure. I think you've got the wrong end of the stick in terms of the templating structure.

You could make  template like this and either include your current one or add the HTML?
`
<ul>
{% for item in value.items.all %}
{% include "your-template-path" with value=item %}
{% endfor %}
</ul>
`

chris-lawton: I guess i'm a touch confused about how this will work, if say, I included 3 icon bullet streamfields in a row would it output the following markup?
```
<ul>
  <li class="icon-bullet"></li> 
</ul>
<ul>
  <li class="icon-bullet"></li> 
</ul>
<ul>
  <li class="icon-bullet"></li> 
</ul>
```
It's not the end of the world but would be neater as:
```
<ul>
  <li class="icon-bullet"></li> 
  <li class="icon-bullet"></li> 
  <li class="icon-bullet"></li> 
</ul>
```

zerolab: See #185 (specifically https://github.com/wagtail/wagtail.org/pull/185/files#diff-3bb733f196978d780770738e86608c68a84a8f77913e1d73278b0dc8f24c3397R356-R376) for the Icon Bullets block definition.

It is list of Icon Bullet blocks, Icon Bullet being what you're temlating/styling here

chris-lawton: Thanks Dan. That sheds some light on it. I've updated the templates now.

PR #189, opened 2022-07-28
Headline

chris-lawton:
Adds light and dark variations of the Headline component, both viewable on a newly created 'streamfields' page viewable under Styleguide [in the Pattern Library](http://localhost:3000/pattern-library/render-pattern/patterns/styleguide/streamfields.html).

Also:
- Adds base icon colours
- Renames icons.scss to _icons.scss
- Fixes incorrect global line height
- Adds white background colour to buttons so they are not transparent when used on dark backgrounds


wagtailio/static/sass/components/_headline.scss
```diff
@@ -0,0 +1,109 @@
...
+        position: absolute;
+        width: 230px;
+        height: 300px;
+        top: 0;
+        right: -40px; // stylelint-disable-line property-disallowed-list
+
+        @include media-query(large) {
+            width: 420px;
+            height: 545px;
+            right: 10%; // stylelint-disable-line property-disallowed-list
```
William-Blackie, line 86: It feels wrong to disable this line, could we use `inset` or remove from the  property-disallowed-list

wagtailio/project_styleguide/templates/patterns/components/streamfields/headline/headline.html
```diff
@@ -0,0 +1,25 @@
...
+            <p class="headline__subheading intro-big">{{ value.subheading }}</p>
+        {% endif %}
+
+        {% if value.intro %}
+            <p class="headline__intro">{{ value.intro }}</p>
+        {% endif %}
+
+        {% if value.cta_url or value.cta_page %}
+            <div class="headline__button-wrap">
+                {% include "patterns/components/buttons/button.html" with arrow=True url=url title=value.cta_text classes="headline__button" %}
```
William-Blackie, line 21: I think that this should be provided by the headline, can we do just {% if value.url %}?

wagtailio/static/sass/components/_headline.scss
chris-lawton: I've enabled the use of it as we use it elsewhere

PR #190, opened 2022-07-29
Teaser

chris-lawton:
Adds teaser component viewable locally [in the PL here](http://localhost:3000/pattern-library/render-pattern/patterns/components/streamfields/teaser/teaser.html)

Also:
- Adds meta component (used on other pages)
- Adds author component (used on other pages)
- Reduces h2 line height on mobile - we don't have a value to reference this but it looked too large
- Add basic responsive image css rules
- Allows overriding of images tags in the pattern library

@William-Blackie since I enabled overriding of images in the PL all the links to components are appearing twice in the sidebar - are you able to pull down and take a look - probably something that'll stand out to you.

### DIFF
--- a/wagtailio/project_styleguide/templates/patterns/components/author/author.yaml
+++ b/wagtailio/project_styleguide/templates/patterns/components/author/author.yaml
@@ -1,0 +1,11 @@
+context:
+  value:
+    author: Meagen Voss
+    author_meta: Wagtail community manager
+    author_image: True
+
+tags:
+  image:
+    value.author_image fill-90x90 class="author__image":
+      raw: |
+        <img src="http://via.placeholder.com/90x90" width="90" height="90" alt="Some alt" class="author__image">
--- a/wagtailio/project_styleguide/templates/patterns/components/author/author.yaml
+++ b/wagtailio/project_styleguide/templates/patterns/components/author/author.yaml
@@ -1,6 +1,6 @@
 context:
   value:
-    author: Meagen Voss
+    author: John Smith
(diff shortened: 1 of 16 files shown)


wagtailio/static/sass/components/_teaser.scss
```diff
@@ -0,0 +1,82 @@
+.teaser {
+    $root: &;
+    opacity: 1;
```
William-Blackie, line 3: Question: is there a reason to use opacity: 1 here?

wagtailio/project_styleguide/templates/patterns/components/author/author.yaml
```diff
@@ -0,0 +1,11 @@
+context:
+  value:
+    author: Meagen Voss
```
William-Blackie, line 3: Nitpick: I know this is in the designs, potentially we should refrain from using real names here.

William-Blackie: Or, only use puns because that would be more fun

William-Blackie:
I've tried for probably too long to work out why the pattern library is loading templates twice, can we split this out to unblock this ticket? It does not affect QA / Review, but does need fixing.

wagtailio/static/sass/components/_teaser.scss
chris-lawton: nope! this is my go to css rule to test that everything is working correctly and must have left it in by accident. Thanks!

PR #193, opened 2022-08-01
Highlight

chris-lawton:
Viewable locally in the pattern library here: http://localhost:3000/pattern-library/render-pattern/patterns/components/streamfields/highlight_block/highlight_block.html

Also:
- Adds ability to override images in the pattern library (this has duplicated 
- Makes image responsive 
- Sets all buttons to be `display: inline-flex` by default (I found myself overriding this enough to warrant changing it)

### DIFF
--- a/wagtailio/project_styleguide/templates/patterns/components/streamfields/highlight_block/highlight_block.html
+++ b/wagtailio/project_styleguide/templates/patterns/components/streamfields/highlight_block/highlight_block.html
@@ -1,6 +1,6 @@
 {% load wagtailcore_tags wagtailimages_tags %}
 
-<div class="highlight{% if value.image_on_right or image_on_right %} highlight--reverse{% endif %} grid">
+<div class="highlight{% if image_on_right %} highlight--reverse{% endif %} grid">
     {% image value.image fill-650x500 class="highlight__image" %}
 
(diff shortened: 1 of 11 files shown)


wagtailio/project_styleguide/templates/patterns/components/streamfields/highlight_block/highlight_block.html
```diff
@@ -0,0 +1,24 @@
+{% load wagtailcore_tags wagtailimages_tags %}
+
+<div class="highlight{% if value.image_on_right or image_on_right %} highlight--reverse{% endif %} grid">
```
William-Blackie, line 3: Suggestion:

I think we could simplify it to `if image_on_right`, this can be set via the block's context, or when including the template. We should try if possible to keep FE logic as simple and non-bound to a model's variables as possible.

William-Blackie (commented):
Looks good, just one small suggestion :+1:

wagtailio/project_styleguide/templates/patterns/components/streamfields/highlight_block/highlight_block.html
chris-lawton: Went through this on a call - thanks Will!

PR #195, opened 2022-08-03
Get Started

chris-lawton:
Adds the 'Get Started' component, [viewable locally in the pattern library here](http://localhost:3000/pattern-library/pattern/patterns/components/get-started/get-started.html).

Also:
- Adds icon link component as this is used elsewhere in the build - [viewable locally here](http://localhost:3000/pattern-library/pattern/patterns/components/icon-link/icon-link.html)
- Refactors all icons to use 1 path so it can be animated (which in turn has reduced the sprite sheet from 300kb to 50kb). Previously the `<use>` tag would output the paths in the shadow dom meaning they could not be manipulated.
- Adds an `icon.html` template which should be used going forward
- Update existing components to use `icon.html` include


wagtailio/project_styleguide/templates/patterns/components/get-started/get-started.html
```diff
@@ -0,0 +1,13 @@
+<div class="get-started grid">
+    <h2 class="get-started__heading heading-one">{{ value.heading }}</h2>
+
+    {% if value.items.all %}
```
William-Blackie, line 4: Note: I'd double check your syntax here, `.all` is used in reference to querysets, here you are referencing a list of blocks so you should be able to just do `{% for item in value.items %}`

wagtailio/static/sass/components/_get-started.scss
```diff
@@ -0,0 +1,53 @@
...
+            margin-bottom: 40px;
+        }
+
+        @include media-query(large) {
+            grid-column: 2 / span 5;
+            margin-bottom: 100px;
+        }
+    }
+
+    &__list {
```
William-Blackie, line 26: Note: I think you need to add a little padding (40px) here, there's a slight offset on the list when compared to the title's indent.
image

wagtailio/static/sass/components/_get-started.scss
```diff
@@ -0,0 +1,53 @@
...
+            row-gap: 50px;
+        }
+
+        @include media-query(large) {
+            grid-column: 2 / span 5;
+            flex-flow: row wrap;
+            row-gap: 100px;
+        }
+    }
+
```
William-Blackie, line 46: Note: The icons are a little large. Not massively off, it would be good to hear if you're aiming for consistency across all of the other components.
image

wagtailio/static/sass/components/_headline.scss
```diff
@@ -73,13 +77,24 @@
         position: absolute;
         width: 230px;
         height: 300px;
-        top: 0;
-        right: -40px;
+        right: -20px;
```
William-Blackie, line 80: Nitpick: It would be nice to keep consistent to either `% or px`, that's probably me just being neurotic but I feel like it's easier to understand the difference between the mobile and the desktop position.

wagtailio/static/sass/components/_hero.scss
```diff
@@ -35,24 +35,41 @@
...
+
+            @include media-query(large) {
+                right: 0;
+                width: 410px;
+                height: 410px;
+            }
+        }
+
+        // Icon overrides
+        .icon__svg-fg {
```
William-Blackie, line 67: Comment: Does this make sense to be here? I feel like this might be better as an alternative style like `icon__svg-bg--eggshell`, what do you think?

William-Blackie (changes requested):

wagtailio/project_styleguide/templates/patterns/includes/svg_sprites.html
```diff
@@ -1,54 +1,199 @@
...
...9.739l-75.543-8.501-.447 3.975 75.542 8.501.448-3.975zm-73.829-7.007l-18.991-74.482-3.876.988 18.991 74.482 3.876-.988zm-19.012-74.557c-2.668-8.99-13.104-12.241-20.718-6.708l2.351 3.236c5.503-4 12.692-1.59 14.532 4.61l3.835-1.138zm-20.8-6.646l-59.524 48.098 2.514 3.111 59.524-48.098-2.514-3.11zm-57.25 47.931l-66.496-39.285-2.034 3.444 66.496 39.285 2.034-3.444zm-66.574-39.329c-8.31-4.42-18.192.252-19.583 9.527l3.956.593c.959-6.397 7.743-9.782 13.749-6.588l1.878-3.532zm-19.593 9.604l-8.441 76.4 3.976.439 8.441-76.4-3.976-.44zm-6.947 74.681l-74.482 18.992.988 3.876 74.482-18.992-.988-3.876zM18.22 136.79c-8.99 2.668-12.241 13.104-6.708 20.718l3.236-2.351c-4-5.503-1.59-12.693 4.61-14.533l-1.138-3.834zm-6.654 20.789l48.158 60.382 3.127-2.494-48.158-60.382-3.127 2.494zm48.006 58.107l-39.345 65.638 3.43 2.057 39.346-65.639-3.431-2.056zm-39.396 65.728c-4.42 8.31.252 18.192 9.527 19.583l.593-3.956c-6.397-.959-9.782-7.743-6.588-13.749l-3.532-1.878zm9.604 19.593l76.4 8.441.439-3.976-76.4-8.441-.44 3.976zm74.681 6.947l18.992 74.482 3.876-.989-18.992-74.481-3.876.988zm19.013 74.556c2.668 8.991 13.104 12.242 20.718 6.709l-2.352-3.236c-5.502 3.999-12.692 1.59-14.532-4.611l-3.834 1.138zm20.801 6.6
```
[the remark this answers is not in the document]
chris-lawton: Good spot! I've removed the class references as the refactor moves away from using that method. Unfortunately the paper plane svg does not follow the format of the other SVGs so it will need more work if we want to be able to animate this one.

wagtailio/project_styleguide/templates/patterns/components/get-started/get-started.html
chris-lawton: thanks, updated!

wagtailio/static/sass/components/_get-started.scss
chris-lawton: Good spot. Added this in and reduced the percentage padding so they don't get too squashed on a laptop sized screen (just before the medium brekapoint kicks in)

wagtailio/static/sass/components/_get-started.scss
chris-lawton: Definitely want to match the design so i've reduced these down. Thanks.

wagtailio/static/sass/components/_headline.scss
chris-lawton: I'm happy to keep them as `px` - feels more predictable to me but happy either way.

wagtailio/static/sass/components/_hero.scss
chris-lawton: I think so. I've standardised all icons to have the [same background colour](https://github.com/wagtail/wagtail.org/blob/10-get-started/wagtailio/static/sass/components/_icon.scss#L13) and [the same stroke](https://github.com/wagtail/wagtail.org/blob/10-get-started/wagtailio/static/sass/components/_icon.scss#L21), so now they can be manipulated on a component-by-component basis. 

The reason I moved away from using colours in class names was because when they change - on hover, for example - [they might turn a different colour](https://github.com/wagtail/wagtail.org/blob/10-get-started/wagtailio/static/sass/components/_icon-link.scss#L58). Having a `foreground` and a `background` class doesn't tie you in to any specifics.

wagtailio/project_styleguide/templates/patterns/includes/svg_sprites.html
```diff
@@ -1,54 +1,199 @@
...
...72 15.72 0 01-11.291 12.948 15.72 15.72 0 01-16.398-5.122h0l-53.715-65.007a15.714 15.714 0 01-.829-18.915h0l41.981-61.099a16.721 16.721 0 00-4.311-23.246 16.718 16.718 0 00-23.245 4.312l-41.981 61.098-.001.001a15.718 15.718 0 01-17.954 6.009l-79.955-26.819a15.72 15.72 0 01-10.653-13.479 15.714 15.714 0 018.052-15.179h0L742.09 55.073h.001a15.717 15.717 0 0123.161 15.915zm-144.38 156.681l41.981-61.099a16.217 16.217 0 1126.732 18.368l-41.981 61.098-26.732-18.367z" stclass="off-black"/><path d="M5.58 522.692c3.936-19.578 10.678-30.173 28.77-42.416 56.933-38.527 129.755-51.816 196.282-61.553 57.104-8.358 113.921-8.516 171.296-11.356 24.015-1.188 52.375-4.478 73.3-17.376 25.027-15.427 44.471-41.926 61.452-65.161 17.616-24.107 34.641-49.92 44.727-78.233 5.308-14.9 7.189-30.521 13.013-45.289" s stroke-width="7" stroke-linecap="round"/><path d="M1 514c4.508-19.455 11.556-29.847 30-41.556 58.037-36.842 131.217-47.993 198-55.777 57.324-6.682 114.122-5.176 171.556-6.334 24.039-.484 52.484-2.942 73.777-15.222 25.468-14.687 45.68-40.605 63.334-63.333 18.315-23.58 36.088-48.883 47-76.889 5.742-14.738 8.08-30.298 14.333-44.889" stroke="#000" stroke-linecap="round" stroke-dasharray="6 6"/></symbol>
```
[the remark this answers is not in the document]
chris-lawton: Thanks, should be resolved now as per the below : )

PR #194, opened 2022-08-03
Backend | ContentPage

engineervix:
This PR does the following:

1. Reorganises **models** in the `core` app, introducing a structure as follows:

   ```console
   .
   ├── models
   │   ├── __init__.py
   │   ├── mixins.py
   │   ├── pages.py
   │   └── snippets.py
   ```

   All the models previously in the `utils` app are moved to the `core` app, with the resulting migrations being modified as follows:

      - [`migrations.SeparateDatabaseAndState`](https://docs.djangoproject.com/en/3.2/ref/migration-operations/#django.db.migrations.operations.SeparateDatabaseAndState) is used to eliminate changes to the database
      - use [`migrations.AlterModelTable`](https://docs.djangoproject.com/en/3.2/ref/migration-operations/#altermodeltable) to rename tables
      - where applicable, set `db_operations` to an empty list to prevent Django from applying any changes to the database

2. Adds a new `ContentStoryBlock` consisting of blocks defined in #185 
3. Adds a new `ContentPage` model in the `core` app, whose `body` StreamField is the `ContentStoryBlock` above
4. Adds a new `HeroMixin`, which is used by the `ContentPage` above, and will be used by other `Page`s

Here are some screenshots:

image

image

PR #196, opened 2022-08-03
Backend | ContentPage

engineervix:
This PR supersedes #194 and does the following:

1. Reorganises **models** in the `core` app, introducing a structure as follows:

   ```console
   .
   ├── models
   │   ├── __init__.py
   │   ├── mixins.py
   │   ├── pages.py
   │   └── snippets.py
   ```

2. Adds a new `ContentStoryBlock` consisting of blocks defined in #185 
3. Adds a new `ContentPage` model in the `core` app, whose `body` StreamField is the `ContentStoryBlock` above
4. Adds a new `HeroMixin`, which is used by the `ContentPage` above, and will be used by other `Page`s

Here are some screenshots:

Screenshot from 2022-08-03 18-07-13

Screenshot from 2022-08-03 18-08-25


zerolab (approved):
Much better! :shipit:

zerolab:
Before I forget, we should get https://github.com/allcaps/wagtail-font-awesome-svg in and register some nice icons to use with all those new blocks (e.g. Cards)

engineervix:
> Before I forget, we should get https://github.com/allcaps/wagtail-font-awesome-svg in and register some nice icons to use with all those new blocks (e.g. Cards)

Thanks, will open a PR for this

PR #198, opened 2022-08-04
Header

chris-lawton:
Adds header containing wagtail logo. I moved the logo to the sprite sheet from static as it'll be called at least twice per page (one header, one footer).

Additional elements that appear in the header are covered in separate tickets (Github star, Navigation, Search, Get Started).


William-Blackie (approved):
Looks good!

PR #203, opened 2022-08-08
Search

chris-lawton:
Adds search dropdown in header [viewable here](http://localhost:8000/pattern-library/render-pattern/patterns/pages/content_page/content_page.html).

Search results are to come in a separate ticket.


package-lock.json
```diff
@@ -38,8 +38,8 @@
     },
```
William-Blackie, line 38: Question: What's changed here in order for us to generate a new lock?

wagtailio/project_styleguide/templates/patterns/includes/header.html
```diff
@@ -5,14 +5,23 @@
...
+            <svg class="header__search-icon" aria-hidden="true">
+                <use xlink:href="#magnifying-glass"></use>
             </svg>
         </button>
-    {% endif %}
+
+        {% if get_started_menu_items %}
+            <button class="header__button header__button--get-started button" data-get-started>
+                Get Started
+                <svg class="modal__close-icon" aria-hidden="true">
```
William-Blackie, line 18: Comment: Does aria-hidden make sense here? The MDN docs [here](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-hidden#description), the quote of ` The aria-hidden attribute can be used to hide non-interactive content from the accessibility API.` makes me think like we don't want it as we can interact with this and the close menu.

wagtailio/project_styleguide/templates/patterns/includes/header.html
```diff
@@ -5,14 +5,23 @@
...
     </a>
 
-    {% if get_started_menu_items %}
-        <button class="header__button header__button--get-started button" data-get-started>
-            Get Started
-            <svg class="modal__close-icon" aria-hidden="true">
-                <use xlink:href="#chevron"></use>
+    <div class="header__items-wrap">
+        <button class="header__button header__button--search" aria-label="Open the search" data-site-search>
+            <svg class="header__search-icon" aria-hidden="true">
```
William-Blackie, line 10: Comment defo think that this aria-hidden might not be correct, please shout if my understanding is wrong

wagtailio/project_styleguide/templates/patterns/includes/search.html
```diff
@@ -0,0 +1,23 @@
...
+            <main class="modal__content">
+                <button class="modal__close-button" aria-label="Close the Search" data-micromodal-close>
+                    <svg class="modal__close-icon" aria-hidden="true">
+                        <use xlink:href="#cross"></use>
+                    </svg>
+                </button>
+                <form action="#" method="get" role="search" class="search">
+                    <div class="search__wrap">
+                        <label class="u-sr-only" for="search-page">Enter a search term</label>
+                        <input class="search__input" id="search-page" type="text" name="query"{% if search_query %} value="{{ search_query }}"{% endif %}>
```
William-Blackie, line 13: image
Looks like parts of a `g` character will get cut off, I think `height: 100%` would sort this

wagtailio/project_styleguide/templates/patterns/includes/search.html
```diff
@@ -0,0 +1,23 @@
+<div class="modal modal--search micromodal-slide" id="site-search-modal" aria-hidden="true">
+    <div class="modal__overlay" tabindex="-1" data-micromodal-close>
+        <div class="modal__container" role="dialog" aria-modal="true">
+            <main class="modal__content">
+                <button class="modal__close-button" aria-label="Close the Search" data-micromodal-close>
```
William-Blackie, line 5: https://user-images.githubusercontent.com/22497837/183664720-c4bcadc3-d22e-424a-bd81-19f17f39cd44.mp4

I think the offset makes the onclick of the close a little inconsistent, is there anything we can do about this?

William-Blackie (changes requested):

package-lock.json
chris-lawton: I think i generated the previous one with an old version of node

wagtailio/project_styleguide/templates/patterns/includes/header.html
chris-lawton: In this instance I see the icon as decorative so it doesn't need to be in accessibility tree - the screen reader will pick up the button text here. From the same article from MDN it sound like this will benefit users of assistive tech:

> This can improve the experience for assistive technology users by hiding:
> - Purely decorative content, such as icons or images

wagtailio/project_styleguide/templates/patterns/includes/header.html
chris-lawton: I think this one's ok based on the guidance from MDN and that the button has an `aria-label`

wagtailio/project_styleguide/templates/patterns/includes/search.html
chris-lawton: Thanks!

wagtailio/project_styleguide/templates/patterns/includes/search.html
chris-lawton: Great spot. Removing the pointer events from the icon seems to remedy this.

chris-lawton:
Thanks for the review @William-Blackie. Think i've resolved all the feedback so let me know if you're happy and we can merge this : )

PR #201, opened 2022-08-04
"Getting started" & "Sign-up form" snippets

engineervix:
This PR builds up on #185 and #196 by 

1. adding two new snippet models -- `GetStartedSnippet` and `SignupFormSnippet`
2. adding two `SnippetChooserBlock`s to the `ContentStoryBlock`, in accordance with the above snippets

In addition, [wagtail-font-awesome-svg](https://github.com/allcaps/wagtail-font-awesome-svg) is added to the project, and the icons for some of the StreamField blocks are updated, as shown in the screenshot below:

Screenshot from 2022-08-04 20-16-45

PR #205, opened 2022-08-09
Minimal Content page template

engineervix:
This PR primarily creates a template (albeit a minimal one, to serve as a starting point) for the **ContentPage** model, introduced in #196. The main purpose is to facilitate QA testing for #185, #196, #201, etc. In addition,

- a couple of templates are also updated (syncing of field names):
  - headline template
  - hero template
- there are a few bugfixes:
  - Fix `ContentStoryBlock`'s template path
  - Specify template in `ContentPage` model
- `CTABlock` is updated to use a [StructValue](https://docs.wagtail.org/en/v2.16.2/advanced_topics/customisation/streamfield_blocks.html#additional-methods-and-properties-on-structblock-values) so that the CTA's `.url` is available on the template


wagtailio/core/blocks.py
```diff
@@ -188,6 +188,13 @@ class Meta:
         template = "core/blocks/home_page_block.html"
 
 
+class CTALinkStructValue(blocks.StructValue):
```
William-Blackie, line 191: Praise: I really like this approach, I suggest to improve this slightly by refactoring existing implementations that handle this logic as well. (Search for usage of the CTABlock model) as we're handling this in context https://github.com/wagtail/wagtail.org/blob/513c9c14323b934cce1e046dc468f5191be5599d/wagtailio/core/blocks.py#L284. We could also add validation for supplying on or the other (or none). What do you think?

wagtailio/core/blocks.py
```diff
@@ -566,4 +574,4 @@ class ContentStoryBlock(blocks.StreamBlock):
     comparison_table = ComparisonTableBlock()
 
     class Meta:
-        template = "patterns/components/streamfields/content_story_block.html"
+        template = "patterns/components/content_story_block.html"
```
William-Blackie, line 577: Suggestion: Could we move the streamfield block into streamfields? I appreciate that this is possibly not the most semantic place but I think the grouping is more logical.

wagtailio/project_styleguide/templates/patterns/components/hero/hero.html
```diff
@@ -1,18 +1,23 @@
...
             {% endif %}
 
             {% if value.intro %}
-                <p class="hero__intro">{{ value.intro }}</p>
+                <p class="hero__intro">{{ value.intro|richtext }}</p>
             {% endif %}
 
-            {% if value.cta_url or value.cta_page %}
-                {% include "patterns/components/buttons/button.html" with arrow=True url=url title=value.cta_text classes="hero__button" %}
+            {% if value.cta %}
```
William-Blackie, line 16: Note: Unless I'm mistaken, you can just you the for loop below for this, if the cta value is `None` the loop will not output anything.

wagtailio/project_styleguide/templates/patterns/components/streamfields/headline/headline.yaml
```diff
@@ -4,7 +4,7 @@ context:
     subheading: Created by the digital agency Torchbox in 2014
     intro: Once Wagtail was released as an open source project, 46 external contributors participated in creating version 1.0. Since then, hundreds of contributors from around the world have contributed to Wagtail. The first Wagtail conference, called Wagtail Space, was held in March 2017 in Arnhem in the Netherlands. The name Wagtail comes from the pretty little birds that gather on the lawns outside Torchbox’s offices from spring to autumn. Torchbox wanted a charming name that had a sense of light and delight to it. Plus, birds make easy logos.
     cta:
-      text: Find out more
+      cta_text: Find out more
```
William-Blackie, line 7: Note: It's not a major issue but the naming convention of this is a little redundant, the prefix is the model name which we already know as it's the block the editor is using.

wagtailio/project_styleguide/templates/patterns/pages/content_page/content_page.html
```diff
@@ -0,0 +1,22 @@
+{% extends "patterns/base_page.html" %}
+{% load wagtailcore_tags wagtailimages_tags static %}
+
+{% block extra_css %}
+{% endblock %}
```
William-Blackie, line 5: Suggestion: These blocks are meant to override the super (base.html / base_page.html) template. As we're not actively doing anything with them we don't need to add them to the template as it's just additional lines for no functional change.

wagtailio/project_styleguide/templates/patterns/pages/content_page/content_page.html
```diff
@@ -0,0 +1,22 @@
...
+
+{% block body_class %}{% endblock %}
+
+{% block content %}
+    {# title #}
```
William-Blackie, line 10: Note: A comment should provide additional context for a developer to understand what's happening, I think `{% if page.title %} provides this context already, what do you think?

William-Blackie: Please note this applies to other comments below, too.

wagtailio/services/templates/services/services_page.html
```diff
@@ -3,7 +3,6 @@
 {% load static wagtailcore_tags wagtailimages_tags %}
 
 {% block content %}
-    {% include "services/includes/svg_sprites.html" %}
```
William-Blackie, line 6: Comment: Should we remove the sprites if we're not using them anymore?

William-Blackie (changes requested):
Hi @engineervix, 

great work! I've left a couple of things for you to look at, feel free to press more for more detail/explanation.

wagtailio/services/templates/services/services_page.html
engineervix: I noticed that the file `services/includes/svg_sprites.html` seems to have been removed in a376628, and replaced with `patterns/includes/svg_sprites.html`, which is already `include`d in the `base.html` template (#181).

wagtailio/project_styleguide/templates/patterns/pages/content_page/content_page.html
engineervix: I agree! Good shout, thanks!

wagtailio/core/blocks.py
engineervix: Sounds great, thanks! TIL more of how to take advantage of mixins!

Updated in 3687292

wagtailio/core/blocks.py
engineervix: Great! Done in d79fba7

wagtailio/project_styleguide/templates/patterns/components/hero/hero.html
engineervix: Thanks, done in ae8d891

wagtailio/project_styleguide/templates/patterns/components/streamfields/headline/headline.yaml
engineervix: Makes sense! Done in ae8d891 and 376e73b

wagtailio/project_styleguide/templates/patterns/pages/content_page/content_page.html
engineervix: Thanks, done in ae8d891

wagtailio/project_styleguide/templates/patterns/pages/content_page/content_page.html
engineervix: updated in ae8d891

wagtailio/services/templates/services/services_page.html
William-Blackie: Good point, thanks @engineervix.

wagtailio/core/blocks.py
```diff
@@ -188,20 +188,25 @@ class Meta:
         template = "core/blocks/home_page_block.html"
 
 
-class CTABlock(blocks.StructBlock):
-    cta_text = blocks.CharBlock(label="CTA text", max_length=255)
-    cta_page = blocks.PageChooserBlock(label="CTA page", required=False)
-    cta_url = blocks.URLBlock(label="CTA URL", required=False)
+class CTALinkMixin(blocks.StructBlock):
+    def is_required(self):
```
William-Blackie, line 192: Praise: Great work here Victor, this is very clean and clear overall, I really appreciate the effort you've put in here!

Small suggestion: I would personally name this to `is_link_required`, is_required doesn't convey to another developer what is actually being required without digging through the code to the superclass. While this is not an issue now (we can just scroll up a few lines!), if we think into the future when this project as a whole gets more complicated it's small details like this that make a project easier to work on.

wagtailio/core/blocks.py
```diff
@@ -226,10 +231,24 @@ def get_context(self, value, parent_context=None):
             context["value"]["url"] = value["cta_page"].get_url
         return context
 
+
+class CTALinkStructValue(blocks.StructValue):
```
William-Blackie, line 235: Suggestion: We can take this one step further, if we move this class above and then assign it to the mixin just like how you've done in the CTABlock model, we can inherit this assignment from the mixin itself, we won't need to re-define it within each model that uses the Mixin. What do you think?

wagtailio/core/blocks.py
```diff
@@ -226,10 +231,24 @@ def get_context(self, value, parent_context=None):
...
+    def url(self):
+        cta_url = self.get("cta_url")
+        cta_page = self.get("cta_page")
+        return cta_url or cta_page.url
+
+
+class CTABlock(CTALinkMixin, blocks.StructBlock):
```
William-Blackie, line 242: Note: Something to be careful of is the Python MRO(Method Resolution Order). As the MRO is left to right, we're inheriting a modified StructBlock(The mixin) and then inheriting the `StructBlock` again from a direct import, we only need the mixin. Take a look at this explainer when you can https://www.geeksforgeeks.org/method-resolution-order-in-python-inheritance/

wagtailio/core/blocks.py
```diff
@@ -245,7 +264,7 @@ class Meta:
         label = "Card"
 
 
-class LogoCardBlock(blocks.StructBlock):
+class LogoCardBlock(CTALinkMixin, blocks.StructBlock):
```
William-Blackie, line 267: Note: We do not need the second inherited model.

wagtailio/core/blocks.py
engineervix: Sounds good! :+1:

wagtailio/core/blocks.py
engineervix: Ah! I see! Many thanks for this :+1:

wagtailio/core/blocks.py
engineervix: Fixed in 9d75fc4

wagtailio/core/blocks.py
engineervix: Fixed in 9d75fc4

wagtailio/core/blocks.py
engineervix: implemented in 9d75fc4

wagtailio/core/blocks.py
engineervix: Makes sense! updated in 128de66

William-Blackie (approved):
Thanks for making those changes @engineervix, looks great.

PR #206, opened 2022-08-10
Blog Page

chris-lawton:
Adds the blog page, viewable [in the pattern library here](http://localhost:8000/pattern-library/render-pattern/patterns/pages/blog/blog_page.html).

Also:
- Adds a `rich-text-example.html` template
- Adds rich text link hover styles (these had been added previously so not sure where they went)
- Adds `related-content.html` which makes use of `logo_card_block.html`

### DIFF
--- a/wagtailio/project_styleguide/templates/patterns/components/related-content/related-content.html
+++ b/wagtailio/project_styleguide/templates/patterns/components/related-content/related-content.html
@@ -1,0 +1,13 @@
+{% if related_pages %}
+    <div class="related-content grid">
+        <h3 class="related-content__heading">You might also like</h3>
+
+        <div class="related-content__items">
+            {% for related in related_pages %}
+                <article class="related-content__item">
+                    {% include "patterns/components/streamfields/cards/logo_card_block.html" with value=related classes="logo-card--article" %}
+                </article>
+            {% endfor %}
+        </div>
+    </div>
+{% endif %}
--- a/wagtailio/project_styleguide/templates/patterns/components/streamfields/cards/logo_card_block.html
+++ b/wagtailio/project_styleguide/templates/patterns/components/streamfields/cards/logo_card_block.html
@@ -1,6 +1,6 @@
 {% load wagtailcore_tags wagtailimages_tags %}
 {% if value.link_url %}
-    <a href="{{ value.link_url }}" class="logo-card">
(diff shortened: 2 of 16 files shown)


wagtailio/project_styleguide/templates/patterns/components/related-content/related-content.html
```diff
@@ -0,0 +1,13 @@
+{% if related_pages %}
+    <div class="related-content grid">
+        <h3 class="related-content__heading">You might also like</h3>
```
chris-lawton, line 3: Not sure if this is hardcoded or not - couldn't find a corresponding field for it.

wagtailio/project_styleguide/templates/patterns/components/streamfields/cards/logo_card_block.html
```diff
@@ -12,9 +12,15 @@ <h4 class="logo-card__heading teaser-heading">{{ value.heading }}</h4>
     <p class="logo-card__description">{{ value.description }}</p>
 {% endif %}
 
-{% if value.logo %}
+{% if value.author and value.logo %}
```
chris-lawton, line 15: `value.logo` for the author avatar doesn't feel right. Maybe it's a wider discussion of if this should be it's own template however they are very similar apart from the logo / avatar

stevedya (approved):
Looks great! Just a couple suggestions, nothing blocking.

wagtailio/project_styleguide/templates/patterns/components/related-content/related-content.html
```diff
@@ -0,0 +1,13 @@
...
+
+        <div class="related-content__items">
+            {% for related in related_pages %}
+                <article class="related-content__item">
+                    {% include "patterns/components/streamfields/cards/logo_card_block.html" with value=related classes="logo-card--article" %}
```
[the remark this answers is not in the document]
chris-lawton: Yeah I agree - I think we took the name from the miro board and then realised it could be used for other things. I'll add a ticket in the backlog to rename : )

wagtailio/project_styleguide/templates/patterns/components/streamfields/cards/logo_card_block.html
chris-lawton: I think this may need [changing in BE too](https://github.com/wagtail/wagtail.org/blob/34-blog-page/wagtailio/core/blocks.py#L274) so i'll raise with them to see what they suggest.

PR #208, opened 2022-08-10
32 comparison table

William-Blackie:
Local dev link [here](http://0.0.0.0:3000/pattern-library/pattern/patterns/components/streamfields/comparison_table_block/comparison_table_block.html)

This implements the comparison table block, please note that I've had to override the pattern library due to how the nesting structure is set in the table and having the override only allowing on block type, meaning we have to approach this in a slightly non-standard way.

Please take a hard look at the mobile view as it technically meets the described criteria but I feel like it could be improved, it would require a fairly drastic change though.


wagtailio/static/sass/components/_comparison-table.scss
```diff
@@ -0,0 +1,42 @@
...
+    tbody {
+        grid-column: 2 / span 1;
+        grid-gap: 20px;
+
+        @include media-query(medium) {
+            grid-column: 2 / span 3;
+        }
+    
+        @include media-query(large) {
+            grid-column: 2 / span 7;
```
chris-lawton, line 17: Suggestion: update this to `grid-column: 2 / span 5` as we only have 5 columns for content at the large breakpoint

wagtailio/static/sass/components/_comparison-table.scss
```diff
@@ -0,0 +1,42 @@
+.comparison-table {
+    align-items: start;
+
+    td:first-child {
+        font-weight: $weight--bold;
+    }
+
+    tbody {
+        grid-column: 2 / span 1;
```
chris-lawton, line 9: suggestion: update this to `grid-column: 2 / span 2;` to span the 2 content columns we have on mobile

wagtailio/static/sass/components/_comparison-table.scss
```diff
@@ -0,0 +1,42 @@
+.comparison-table {
+    align-items: start;
```
chris-lawton, line 2: question: was this added for text alignment? I usually see it associated with `display: flex` so i'm curious what effect it was with that property added

wagtailio/static/sass/components/_comparison-table.scss
```diff
@@ -0,0 +1,42 @@
+.comparison-table {
```
chris-lawton, line 1: Suggestion: add some space under this component so the next component doesn't sit too close to it. There's 100px at the large breakpoint and then i've been reducing that down to 20px for mobile.

image

wagtailio/static/sass/components/_comparison-table.scss
```diff
@@ -0,0 +1,42 @@
...
+
+    td {
+        padding: 20px 20px 20px 0;
+        border: 1px $color--light-grey;
+        border-style: none none dashed none;
+
+        > div > p {
+            font-size: 1rem;
+            line-height: 1.5rem;
+            @include media-query(large) {
```
chris-lawton, line 36:
```suggestion
            line-height: 1.5rem;
            
            @include media-query(large) {
```

wagtailio/static/sass/components/_comparison-table.scss
```diff
@@ -0,0 +1,42 @@
+.comparison-table {
+    align-items: start;
+
+    td:first-child {
+        font-weight: $weight--bold;
+    }
+
+    tbody {
```
chris-lawton, line 8: suggestion: I think we should allow scrolling on the table only when the table is wider than the page - currently the whole page scrolls. It can be achieved with by adding the following:

```
    tbody {
        grid-column: 2 / span 2;
        grid-gap: 20px;
        overflow-x: scroll; // new

        @include media-query(medium) {
            grid-column: 2 / span 3;
            overflow-x: auto; // new
        }

```

chris-lawton (changes requested):
Just a few tweaks and this one should be good to go

wagtailio/project_styleguide/templates/patterns/_pattern_library_only/comparison_table.html
```diff
@@ -0,0 +1,122 @@
+{# This template provides data for the comparison table, to update it use the streamfield HTML #}
+<tr>
+    <td>
+    </td>
+
+    <td>
```
chris-lawton, line 6: Suggestion: these should be `<th>` if they are part of table header. Can this be added via the table streamfield?

William-Blackie: Oh dear, I don't think this BE will meet the FE implementation needs

William-Blackie: @zerolab, what do you think, the "heading" block is part of the TypedTableBlock which contains <th>. In the design, we use an SVG / image.

zerolab: I feel the first row should be `th`s (so that would include the logos)

wagtailio/static/sass/components/_comparison-table.scss
William-Blackie: It is, I'm not sure why I left this, my bad.

wagtailio/static/sass/components/_comparison-table.scss
William-Blackie: Neat!

wagtailio/project_styleguide/templates/patterns/_pattern_library_only/comparison_table.html
William-Blackie: Thanks both.

PR #209, opened 2022-08-10
Logo card alignment

chris-lawton:
Ensure the logo appears at the same point regardless of how much text each card has


William-Blackie:
Looks good @chris-lawton!

PR #212, opened 2022-08-11
FeatureIndexPage

engineervix:
The FeatureIndexPage is revised to match updated needs:
- new fields are introduced -- **subheading**, **cta** & **get_started**
- the **body** StreamField is updated to use the new `FeatureIndexPageBlock`,
  a StreamBlock consisting of `FeatureCategoryBlock`s
- use new template -- `patterns/pages/feature_index_page/feature_index_page.html`

The intention is to replace the existing [`features`](https://github.com/wagtail/wagtail.org/tree/d4bd1bb901c45803147290fd60abf254338fce6a/wagtailio/features) app

<details><summary>Screenshots</summary>

**Wagtail Admin**

Screenshot from 2022-08-18 20-33-54

**Minimal Template**

Screenshot from 2022-08-18 20-39-41

</details>


wagtailio/features/blocks.py
```diff
@@ -39,9 +39,37 @@ class Meta:
...
 
 
+class FeatureBlock(blocks.StructBlock):
+    heading = blocks.CharBlock(label="Feature heading", required=False, max_length=255)
+    description = blocks.RichTextBlock(
+        label="Feature description",
+        required=True,
+        features=["bold", "italic", "link", "document"],
+    )
+    link = blocks.URLBlock(label="Feature link", required=False)
```
William-Blackie, line 49: Note: I think we'll need another field here. The design has "view docs" which look like it could be hard-coded but what happens if we're not linking to the docs?

William-Blackie: To be clear, we'll need a button title field

wagtailio/features/blocks.py
```diff
@@ -39,9 +39,37 @@ class Meta:
...
+
+class FeatureCategoryBlock(blocks.StructBlock):
+    heading = blocks.CharBlock(required=False, max_length=255)
+    features = blocks.ListBlock(FeatureBlock(), min_num=2)
+
+    class Meta:
+        icon = "folder-inverse"
+        label = "Feature Category"
+        template = (
+            "patterns/components/streamfields/features/feature_category_block.html"
```
William-Blackie, line 65: Suggestion: I think it might be cleaner to use one template, and not provide templates to the child templates, instead of creating multiple nested-child templates. What do you think?

wagtailio/project_styleguide/templates/patterns/pages/feature_index_page/feature_index_page.html
```diff
@@ -0,0 +1,10 @@
+{% extends "patterns/base_page.html" %}
+{% load wagtailcore_tags wagtailimages_tags static %}
+
+{% block content %}
+    <h3>{{ page.subheading }}</h3>
```
William-Blackie, line 5: Note: subheading can be blank; just double-check you've added the correct handling of None.

wagtailio/project_styleguide/templates/patterns/pages/feature_index_page/feature_index_page.html
```diff
@@ -0,0 +1,10 @@
...
+    {% for category in page.feature_categories %} ☐ {{ category }} {% endfor %}
```
William-Blackie, line 6: I feel like this should be handled from the block itself, the logic allows calls to the child blocks so should be dependent on that block. What do you think?

wagtailio/project_styleguide/templates/patterns/components/streamfields/features/feature_block.html
```diff
@@ -0,0 +1,7 @@
+{% load static wagtailcore_tags %}
+
+<div>
+    <h4>{{ value.heading }}</h4>
+    <div class="rich-text">{{ value.description|richtext }}</div>
+    <a href="{{ value.link }}">View docs</a>
```
William-Blackie, line 6: I could be wrong, but do you think we should allow this to be editable?

William-Blackie (changes requested):

William-Blackie:
Great work @engineervix, just some nitpicks / requests for you.

wagtailio/features/blocks.py
engineervix: Thanks @William-Blackie 

Are you referring to 

1. dropping the template definition for the **FeatureBlock**?

https://github.com/wagtail/wagtail.org/blob/7b09edd312419bd3bf96efea71d7033a2fb78b87/wagtailio/features/blocks.py#L54

  and instead having the markup defined in `patterns/components/streamfields/features/feature_category_block.html` ?
 
2. or  perhaps modifying `patterns/components/streamfields/features/feature_category_block.html` as follows:

```diff
<div>
    <h3>{{ value.heading }}</h3>
    {% for feature in value.features %}
-        {% include "patterns/components/streamfields/features/feature_block.html" with value=feature %}
+        {% include_block feature with value=feature %}
    {% endfor %}
</div>
```

wagtailio/project_styleguide/templates/patterns/components/streamfields/features/feature_block.html
engineervix: As indicted in https://github.com/wagtail/wagtail.org/pull/212#discussion_r946862957, I think it makes sense to allow for flexibility for editors, while still providing a default / fallback value (**View docs**)

```suggestion
    {% if value.link %}
        <a href="{{ value.link }}">
            {% if value.link_title %}{{ value.link_title }}{% else %}View docs{% endif %}
        </a>
    {% endif %}
```

wagtailio/project_styleguide/templates/patterns/pages/feature_index_page/feature_index_page.html
engineervix: Screenshot from 2022-08-18 19-25-22

The idea behind this line was to be able to list all the headings for the available feature categories, based on the Figma design in the image above (with a view to using JavaScript [via corresponding FE ticket] to perform the filtering). Is there a way this can be achieved on the block itself?

wagtailio/features/blocks.py
engineervix: Thanks, updated in 1a36be7

wagtailio/features/blocks.py
engineervix: I've used the second approach in 81669e0, but happy to update if needed.

wagtailio/project_styleguide/templates/patterns/pages/feature_index_page/feature_index_page.html
engineervix: Thanks, handled in 99b69e6

wagtailio/project_styleguide/templates/patterns/components/streamfields/features/feature_block.html
engineervix: implemented in 1a36be7

William-Blackie: hmm, I'm not sure if I agree, what happens if someone does not have anything meaningful to link to for a single point?

William-Blackie: The more flexible solution would be to either, supply a link title or be blank?

engineervix: > hmm, I'm not sure if I agree, what happens if someone does not have anything meaningful to link to for a single point?

If a link is not provided, then we do not worry about the link title, and the anchor element is not rendered

engineervix: > The more flexible solution would be to either, supply a link title or be blank?

What would be the best way to handle a situation where the link title is not specified? My thoughts: since we probably don't want to have something like `<a href="{{ value.link }}"></a>`, we could

1. `<a href="{{ value.link }}">{{ value.link }}</a>`  -- the problem is the link may be too long
2.  `<a href="{{ value.link }}">{{ value.link|cut:"https://"|cut:"http://"|cut:"www." }}</a>`
3. update context in **FeatureBlock** so that we have a "trimmed link" with no scheme and www subdomain, then use this as link title if one isn't provided

wagtailio/project_styleguide/templates/patterns/pages/feature_index_page/feature_index_page.html
```diff
@@ -0,0 +1,10 @@
+{% extends "patterns/base_page.html" %}
+{% load wagtailcore_tags wagtailimages_tags static %}
+
+{% block content %}
+    {% if page.subheading %}<h3>{{ page.subheading }}</h3>{% endif %}
+    {% for category in page.feature_categories %} ☐ {{ category }} {% endfor %}
+    {% include_block page.body %}
+    {% include_block page.cta %}
```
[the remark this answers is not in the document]
engineervix: indeed, makes sense to use `cta.html`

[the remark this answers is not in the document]
engineervix: Fixed in 302c294

wagtailio/project_styleguide/templates/patterns/pages/feature_index_page/feature_index_page.html
```diff
@@ -0,0 +1,12 @@
+{% extends "patterns/base_page.html" %}
+{% load wagtailcore_tags wagtailimages_tags static %}
+
+{% block content %}
+    {% if page.subheading %}<h3>{{ page.subheading }}</h3>{% endif %}
+    {% for block in page.features %} ☐ {{ block.value.heading }} {% endfor %}
```
William-Blackie, line 6: Note: Could this not be done with the features template?

William-Blackie (changes requested):

wagtailio/features/blocks.py
```diff
@@ -39,9 +39,44 @@ class Meta:
...
+        label="Feature description",
+        required=True,
+        features=["bold", "italic", "link", "document"],
+    )
+    link = blocks.URLBlock(label="Feature link", required=False)
+    link_title = blocks.CharBlock(
+        label="Feature link title",
+        required=False,
+        max_length=50,
+        help_text="If left blank, 'View docs' will be used as the default Feature link title",
```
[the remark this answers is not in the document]
engineervix: Thanks, this is removed in 559d2dc

wagtailio/project_styleguide/templates/patterns/pages/feature_index_page/feature_index_page.html
engineervix: Thanks for the guidance @William-Blackie! I've made some revisions in 7a147be and 35d6274

wagtailio/project_styleguide/templates/patterns/components/streamfields/features/feature_category_block.html
```diff
@@ -0,0 +1,16 @@
...
+<div>
+    <h3>{{ value.heading }}</h3>
+    {% comment %}
+    The line below could have been written (for better readability) as
+
+        {% for feature in value.features with value=feature %}
+
+    However, this would break django-pattern library due to a limitation:
+    https://github.com/torchbox/django-pattern-library/issues/138
+    {% endcomment %}
```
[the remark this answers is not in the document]
engineervix: Comment has been removed as it isn't really necessary here.

PR #213, opened 2022-08-11
Add taxonomy app

engineervix:
This PR adds a new taxonomy app to the project, with a taxonomy modeladmin menu item and categories sub-menu item


wagtailio/taxonomy/wagtail_hooks.py
```diff
@@ -0,0 +1,21 @@
...
+from wagtailio.taxonomy.models import Category
+
+
+class CategoryModelAdmin(ModelAdmin):
+    model = Category
+    menu_icon = "tag"
+
+
+class TaxonomiesModelAdminGroup(ModelAdminGroup):
+    menu_label = "Taxonomies"
```
zerolab, line 16: nitpick: 
```suggestion
    menu_label = "Taxonomy"
```

wagtailio/taxonomy/tests.py
```diff
@@ -0,0 +1 @@
+# Create your tests here.
```
William-Blackie, line 1: Suggestion: If we're not using them, do we need this files here?

wagtailio/taxonomy/wagtail_hooks.py
engineervix: Thanks, updated in 74d3e24

wagtailio/taxonomy/tests.py
engineervix: Thanks, removed in 70deda8

PR #215, opened 2022-08-12
Fix/cta text required

William-Blackie:
This fixes the implied functionality of using the CTABlock model where required=False is now respected.

Changes: 
- Template checks for the value of the title as value.cta will return True regardless.
- Set the CTABlock text to required=False and validate when it's conditionally required.


wagtailio/core/blocks.py
```diff
@@ -196,24 +196,17 @@ def url(self):
 
 
 class CTALinkMixin(blocks.StructBlock):
-    def is_link_required(self):
```
William-Blackie, line 199: We actually have the required value passed to Streamfield, we can use that instead.

wagtailio/core/blocks.py
```diff
@@ -196,24 +196,17 @@ def url(self):
...
         errors = {}
         struct_value = super().clean(value)
-
-        if self.is_link_required():
-            if not value.get("cta_page") and not value.get("cta_url"):
-                error = ErrorList(
-                    [ValidationError("You must specify CTA page or CTA URL.")]
-                )
-                errors["cta_url"] = errors["cta_page"] = error
+        if self.required and not value.get("cta_page") and not value.get("cta_url"):
```
William-Blackie, line 205: This is a simplification of the above.

wagtailio/core/blocks.py
```diff
@@ -225,9 +218,12 @@ def clean(self, value):
             )
             errors["cta_url"] = errors["cta_page"] = error
 
+        if not value.get("text") and (value.get("cta_page") or value.get("cta_url")):
```
William-Blackie, line 221: Require text, if someones added a link.

zerolab (approved):
LGTM.

Had a small nitpick on the unnecessary `error` variable assignment for the empty text field check, but it is just that, so let's get this moving :shipit:

PR #216, opened 2022-08-12
Content page

chris-lawton:
Some general Streamfield fixes now the content page BE has been done. 

I've added a `sf-spacing` class so the spacing between streamfields should be consistent.

### DIFF
--- a/wagtailio/project_styleguide/templates/patterns/components/streamfields/rich_text_block/rich_text_block.html
+++ b/wagtailio/project_styleguide/templates/patterns/components/streamfields/rich_text_block/rich_text_block.html
@@ -1,5 +1,7 @@
 {% load wagtailcore_tags %}
 
-<div>
-    {{ value.rich_text }}
+<div class="grid">
+    <div class="rich-text rich-text--sf">
+        {{ value.rich_text }}
+    </div>
 </div>
--- a/wagtailio/project_styleguide/templates/patterns/components/streamfields/rich_text_block/rich_text_block.html
+++ b/wagtailio/project_styleguide/templates/patterns/components/streamfields/rich_text_block/rich_text_block.html
@@ -2,6 +2,6 @@
 
 <div class="grid">
     <div class="rich-text rich-text--sf">
-        {{ value.rich_text }}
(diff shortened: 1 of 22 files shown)


wagtailio/project_styleguide/templates/patterns/components/streamfields/rich_text_block/rich_text_block.html
```diff
@@ -1,5 +1,7 @@
 {% load wagtailcore_tags %}
 
-<div>
-    {{ value.rich_text }}
+<div class="grid">
+    <div class="rich-text rich-text--sf">
+        {{ value.rich_text }}
```
stevedya, line 5: Question: Do we need the `|richtext` filter on here?

wagtailio/static/sass/abstracts/_mixins.scss
```diff
@@ -28,3 +28,12 @@
     white-space: nowrap;
     border: 0;
 }
+
+// Streamfield spacing
+@mixin sf-spacing() {
```
stevedya, line 33: **Praise:** Noice! Definitely needed this

stevedya: We could use this as a block spacer throughout other parts of the site as well (not just for streamfields). Perhaps in the future we can add params for specifying top or bottom spacing or both for different scenarios. Not required as part of this work though 😅

stevedya (approved):
Looking great! Love the addition of the sf-spacing mixin

wagtailio/project_styleguide/templates/patterns/components/streamfields/rich_text_block/rich_text_block.html
chris-lawton: I think we do! : D

wagtailio/static/sass/abstracts/_mixins.scss
chris-lawton: That sounds useful. The streamfields had pretty consistent spacing in the design but no reason why we couldn't rename and adopt this for other components - would be nice to have it all controlled from one source : )

PR #220, opened 2022-08-15
Update icon list

chris-lawton:
- Remove light/dark iterations of icons as there's only 1 version of them now
- Remove arrow from icon choice list
- Add Magnifying Glass and Code Document to list
- Fix paper plane icon
- Add css for paper plane icon
- Use a percentage to make stroke width on icons consistent


wagtailio/core/choices.py
```diff
@@ -46,8 +46,7 @@ class SVGIcon(models.TextChoices):
     ROLLBACK = "rollback", "Rollback"
     SLACK = "slack", "Slack"
     SPEECH_BUBBLE = "speech-bubble", "Speech Bubble"
-    SUN_CLOUD_DARK = "sun-cloud-dark", "Sun Cloud Dark"
-    SUN_CLOUD_LIGHT = "sun-cloud-light", "Sun Cloud Light"
+    SUN_CLOUD = "sun-cloud", "Sun Cloud"
```
chris-lawton, line 49: @engineervix Just FYI i've updated the list of icons after we removed the light/dark versions and added some that were missing

wagtailio/core/choices.py
```diff
@@ -2,7 +2,6 @@
 
 
 class SVGIcon(models.TextChoices):
-    ARROW = "arrow", "Arrow"
```
William-Blackie, line 5: Question: Did you mean to remove this?

William-Blackie: I've just re-read your MR. You did infact mean to remove this :joy:

wagtailio/core/migrations/0045_update_icons_list.py
```diff
@@ -0,0 +1,1259 @@
+# Generated by Django 3.2.12 on 2022-08-19 09:11
```
William-Blackie, line 1: This is a massive nitpick that I feel I should apologise for asking you to rename this slightly. I prefer naming migrations to directly map to a model and what you've changed to generate or the impact. In this case, you've changed the class `SVGIcon` and as it's inheriting from `model.text_choices` I would think semantically the name of `update_svg_icon_choices` as a name is more accurate.

William-Blackie (approved):
Looks great @chris-lawton, nice work! One very small nitpick which I wouldn't even mind you not doing.

wagtailio/core/migrations/0045_update_icons_list.py
chris-lawton: Renamed - thanks Will!

PR #222, opened 2022-08-16
Update BlogPage Model

engineervix:
This PR updates the **BlogPage** model by:

- altering the `body` StreamField to use a newly created `BlogStoryBlock`, which extends the existing `StoryBlock` by adding a `StandAloneCTABlock` introduced in #185 
- adding a _category_, based on #213 
- adding _related pages_

image


zerolab:
#213 being merged, this could use a rebase.

wagtailio/blog/models.py
```diff
@@ -1,22 +1,41 @@
...
 from wagtail_airtable.mixins import AirtableMixin
 from wagtail_content_import.models import ContentImportMixin
 
-from wagtailio.utils.blocks import StoryBlock
+from wagtailio.blog.blocks import BlogStoryBlock
 from wagtailio.utils.mappers import StreamFieldMapper
 from wagtailio.utils.models import CrossPageMixin, SocialMediaMixin
 
 
+class BlogPageCategory(models.Model):
```
zerolab, line 27: Is the plan to potentially allow more than one category? if so, it could use a note.
Otherwise a simple `ForeignKey` to `taxonomy.Category` would do

engineervix: Thanks, updated in 127f2be to use a `ForeignKey` to `taxonomy.Category`

wagtailio/blog/models.py
```diff
@@ -124,6 +157,22 @@ def siblings(self):
...
+    def related_pages(self):
+        return self.page_related_pages.all()
+
+    @cached_property
+    def meta_text(self):
+        return self.category.title
+
+    @cached_property
+    def meta_icon(self):
+        return self.category.icon
```
zerolab, line 170: question: `self.category` can be `None`. what happens when you access a blog post with no category?

wagtailio/blog/blocks.py
```diff
@@ -0,0 +1,6 @@
+from wagtailio.core.blocks import StandaloneCTABlock
+from wagtailio.utils.blocks import StoryBlock
+
+
+class BlogStoryBlock(StoryBlock):
+    standalone_cta = StandaloneCTABlock()
```
zerolab, line 6: question: `StoryBlock` still uses the old streamfield template (https://github.com/wagtail/wagtail.org/blob/127f2be58f11fbac39907d0d3aa80181e62afcf9/wagtailio/utils/blocks.py#L197), and some of its blocks, like the backer blocks reference templates from the blog app ([example](https://github.com/wagtail/wagtail.org/blob/127f2be58f11fbac39907d0d3aa80181e62afcf9/wagtailio/utils/blocks.py#L159)). Will the FE ticket cover this? if not, can you update the relevant paths?

wagtailio/blog/models.py
```diff
@@ -56,6 +63,17 @@ def serve(self, request):
     )
 
 
+class BlogPageRelatedPage(Orderable):
+    parent = ParentalKey("blog.BlogPage", related_name="page_related_pages")
```
zerolab, line 67: The AC was a bit unclear in that it says "As a user I can see related pages (max two) based on category and publish date.". 
I think this is the right approach.

suggestion: considering this is very specific, let's call it as such. i.e. `related_posts`

zerolab (changes requested):
Nice work. Left a couple of suggestions and questions

wagtailio/blog/models.py
engineervix: Good shout! Fixed in f225dad. Also updated `meta_text` accordingly

wagtailio/blog/models.py
engineervix: Thanks, updated in f225dad

wagtailio/blog/blocks.py
engineervix: Need to counter-check with @chris-lawton / @William-Blackie on this one

zerolab: this is the last bit. everything else LGTM as is.
Note: we may want to revisit the blog post streamfield blocks in the future, but that will do for now

William-Blackie: Hey @engineervix, 

I'll reply in here instead of on Slack, I would just change it here to the correct template, even if it's done elsewhere in another MR there's no clash and will help with a review (it can't be reviewed if we can't see it!). 

Thanks - Will.

engineervix: Thanks @zerolab & @William-Blackie. I have updated the template references via 230312a. I've also added the block templates in `wagtailio/utils/blocks.py` to the Pattern Library via ba9ab50

William-Blackie: Hey @engineervix, great work! As we're migrating over to a new layout for templates (which you've gone ahead and done), the logical next step would be to remove the old ones we no longer reference in this directory [here](https://github.com/wagtail/wagtail.org/blob/6f9e6e556e24f1ff13d70157802212faca7c36a7/wagtailio/blog/templates/blog/blocks/image_backer.html).

We'll also need to style these blocks as it looks like it's been removed (most likely when Chris and I initially tooled the project up). Just raise this in standup and we can deal with it.

Thanks!

engineervix: Thanks @William-Blackie 

Done in 423bd9b

PR #229, opened 2022-08-18
Logos

chris-lawton:
Adds FE for logo block viewable [in the pattern library here](http://localhost:8000/pattern-library/pattern/patterns/components/streamfields/logo_block/logo_block.html)

Also fixes the Quotes SF when viewed on the streamfields template in the pattern library

### DIFF
--- a/wagtailio/static/sass/components/_logos.scss
+++ b/wagtailio/static/sass/components/_logos.scss
@@ -1,0 +1,38 @@
+.logos {
+    @include sf-spacing();
+
+    &__heading {
+        margin-bottom: 20px;
+        grid-column: 2 / span 2;
+
+        @include media-query(medium) {
+            margin-bottom: 40px;
+            grid-column: 2 / span 3;
+        }
+
+        @include media-query(large) {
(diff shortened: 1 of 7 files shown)


wagtailio/static/sass/components/_logos.scss
```diff
@@ -0,0 +1,38 @@
...
+            grid-column: 2 / span 4;
+        }
+    }
+
+    &__list {
+        display: flex;
+        flex-wrap: wrap;
+        list-style: none;
+        grid-column: 2 / span 2;
+        gap: 20px;
```
William-Blackie, line 23: All looks good @chris-lawton! I think that maybe we could centre the list in mobile view, it leaves a little too much space for my taste on the right, but feel free to ignore.

William-Blackie (approved):

wagtailio/static/sass/components/_logos.scss
chris-lawton: I agree, think the white space on the right looks odd. Let 's centre them : )

PR #230, opened 2022-08-18
GitHub star button

chris-lawton:
Adds Github star count button viewable on the homepage locally

### DIFF
--- a/wagtailio/project_styleguide/templates/patterns/base_page.html
+++ b/wagtailio/project_styleguide/templates/patterns/base_page.html
@@ -135,8 +135,5 @@
         {% endif %}
 
         <script src="https://platform.twitter.com/widgets.js" async></script>
-        {% block extra_js %}
-            <script async defer src="https://buttons.github.io/buttons.js"></script>
-        {% endblock %}
     </body>
 </html>
--- a/wagtailio/project_styleguide/templates/patterns/base_page.html
+++ b/wagtailio/project_styleguide/templates/patterns/base_page.html
@@ -135,6 +135,8 @@
         {% endif %}
 
         <script src="https://platform.twitter.com/widgets.js" async></script>
-        <script async defer src="https://buttons.github.io/buttons.js"></script>
+        {% block extra_js %}
+            <script async defer src="https://buttons.github.io/buttons.js"></script>
+        {% endblock %}
     </body>
 </html>
--- a/wagtailio/project_styleguide/templates/patterns/includes/header.html
+++ b/wagtailio/project_styleguide/templates/patterns/includes/header.html
@@ -7,7 +7,7 @@
         </a>
 
         <div class="header__github-button">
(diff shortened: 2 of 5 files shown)


wagtailio/project_styleguide/templates/patterns/base_page.html
```diff
@@ -135,6 +135,8 @@
         {% endif %}
 
         <script src="https://platform.twitter.com/widgets.js" async></script>
-        <script async defer src="https://buttons.github.io/buttons.js"></script>
+        {% block extra_js %}
+            <script async defer src="https://buttons.github.io/buttons.js"></script>
+        {% endblock %}
```
William-Blackie, line 140: Thought: I'm not sure we want to place this within the `{% block extra_js %}` as when this is used on other page templates, it will override the super (this one). As this is going to be in the header of every page I think it still makes sense for it to be where it was. What do you think?

wagtailio/project_styleguide/templates/patterns/includes/header.html
```diff
@@ -1,11 +1,18 @@
...
-    </a>
+    <div class="header__col-one">
+        <a class="header__link" href="/" aria-label="Wagtail.org homepage">
+            <svg class="header__logo" aria-hidden="true">
+                <use xlink:href="#wagtail" />
+            </svg>
+        </a>
+
+        <div class="header__github-button">
+            <a class="github-button" href="https://github.com/wagtail/wagtail" data-show-count="true" aria-label="Star Wagtail on GitHub">Star</a>
```
William-Blackie, line 10: Question: This just links straight to the wagtail Github, does the aria-label make sense to screen readers here?

chris-lawton: It was provided by Github but I agree that it needs context. I'll add

wagtailio/project_styleguide/templates/patterns/base_page.html
William-Blackie: So I would say we don't want to add it to the base.html, we use that for the pattern-library and potentially in other non-page-based views like 404 and 500. A good place would be the header but that would suffer the same problem of being able to override it!

chris-lawton: The advice from https://buttons.github.io/ is to `<!-- Place this tag in your head or just before your close body tag. -->`. I personally prefer the latter as that's where I'd expect to find additional scripts that haven't been bundled but I guess the `defer` attribute won't hurt us if we place it in the header.

chris-lawton: > we use that for the pattern-library
Any reason why we wouldn't want to show this in the PL?

chris-lawton: Going to leave it here for the time being - we can always move it done the line if needs be.

PR #233, opened 2022-08-19
Comparison table column fix

William-Blackie:
This tweak forces the table "categories" to maintain their sizing regardless of large images being added.

### DIFF
--- a/wagtailio/static/sass/components/_comparison-table.scss
+++ b/wagtailio/static/sass/components/_comparison-table.scss
@@ -7,7 +7,11 @@
 
     td:first-child {
         font-weight: $weight--bold;
-        min-width: fit-content;
+        width: 150px;
+
+        @include media-query(large) {
+            width: 250px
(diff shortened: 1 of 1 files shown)


wagtailio/static/sass/components/_comparison-table.scss
```diff
@@ -7,6 +7,7 @@
 
     td:first-child {
         font-weight: $weight--bold;
+        min-width: fit-content;
```
chris-lawton, line 10: `min-width: fit-content;` doesn’t have any affect for me. I’d be tempted to set a fixed width of `width: 250px` at the large breakpoint and `width: 150px` on the small breakpoint.

William-Blackie: It would depend on the size of your images in the header and if they pushed the line passed the point of the min-width of the other I guess. Will use your suggestion.

PR #235, opened 2022-08-22
Skip link

chris-lawton:
Adds skip link viewable [in the pattern library here](http://localhost:8000/pattern-library/pattern/patterns/components/skip-link/skip-link.html) - you'll need to tab inside the iframe to see it.

### DIFF
--- a/wagtailio/project_styleguide/templates/patterns/components/skip-link/skip-link.html
+++ b/wagtailio/project_styleguide/templates/patterns/components/skip-link/skip-link.html
@@ -1,0 +1,1 @@
+<a href="#main-content" class="skip-link">Skip to main content</a>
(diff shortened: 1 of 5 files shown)


wagtailio/project_styleguide/templates/patterns/components/skip-link/skip-link.html
```diff
@@ -0,0 +1 @@
+<a href="#main-content" class="skip-link">Skip to main content</a>
```
William-Blackie, line 1: Question: As this is aimed at screen readers, do we want to provide anything further for them? like aria label describing what will happen.

William-Blackie (approved):

wagtailio/project_styleguide/templates/patterns/components/skip-link/skip-link.html
chris-lawton: I would have thought that text within the anchor link would be enough for them to determine what's going to happen. I can't [see any mention](https://webaim.org/techniques/skipnav/) of using an aria label so think this should suffice.

PR #221, opened 2022-08-16
Draft:35- blog index

stevedya:
Addresses monday ticket #35 for the styling of the blog index page.

PR #236, opened 2022-08-22
backend - Blog index page

engineervix:
This Pull Request adds featured posts to the `BlogIndexPage` model, and accordingly updates the templates (defined in #221). 

- [X] a max of 5 `BlogPage`s can be added as featured posts
- [X] the featured posts can only be displayed on the first page, with no filters
- [X] the featured posts are not duplicated

The existing `BlogIndexPage` model is reused, and the context is updated to include `featured_posts` and `categories`.

<details>

  ## wagtail admin
 
 image

 ##  frontend

  image

  ## where page is not the first page

  image

  ## first page but with filter

  image

</details>


wagtailio/blog/models.py
```diff
@@ -55,9 +73,23 @@ def serve(self, request):
             {
                 "page": self,
                 "posts": posts,
+                "featured_posts": [post.page for post in self.featured_posts.all()],
+                "categories": Category.objects.all()
+                .values_list("pk", "title")
+                .distinct()
```
zerolab, line 79: note: distinct is not needed when you query all objects.
suggestion: it would probably be nicer if we returned only categories in use on by `BlogPage`

wagtailio/blog/templates/blog/blog_index_page.html
```diff
@@ -16,7 +16,8 @@ <h1 aria-hidden="true">Latest Chirps</h1>
 
         <section class="blog-listing">
             {% for post in posts %}
-                {% if forloop.first or forloop.counter|divisibleby:6 %}
+                {# only show featured posts on the first pagination page, no filters #}
+                {% if posts.number == 1 and post in featured_posts and 'category' not in request.GET %}
```
zerolab, line 20: question: the designs show featured items every few posts.
looking at this, it just loops through all blog posts (for the first page) and marks it as featured, if it is in featured posts..

This means that if the last 5 blog posts are also marked as featured, they all appear as such. Worth double checking with @vossisboss, but in my head we should have a separation between latest blog posts and featured blog posts, even if it may happen that the featured ones duplicate the latest ones on the first page

zerolab (changes requested):
Got a couple of comments. This is starting to shape up, @engineervix

wagtailio/blog/templates/blog/blog_index_page.html
vossisboss: I think @zerolab is right @engineervix. Featured blog posts will likely be some of our most recent posts, but not always. Sometimes there will be older content we'll want to highlight. I'm fine with the featured ones duplicating the latest ones on the page because the designs are visually distinct. If they all looked the same, then it would be more of an issue. Let me know if that helps @engineervix.

wagtailio/blog/models.py
engineervix: Thanks, updated in cf9286b

wagtailio/blog/templates/blog/blog_index_page.html
zerolab: @engineervix this still need addressing, and then we'd be good to go

engineervix: Thanks @zerolab and @vossisboss, this has been addressed in acfe1c7

With this approach, we'll have duplicate entries on the first page for any of the latest posts which are featured.

wagtailio/project_styleguide/templates/patterns/pages/blog/blog_index_page.html
```diff
@@ -9,8 +9,41 @@
...
             {% for post in posts %}
-                {% if forloop.first or forloop.counter|divisibleby:6 %}
-                    {% include 'patterns/pages/blog/includes/blog_post.html' with post=post featured=True %}
+                {# only show featured posts on the first pagination page, no filters #}
+                {% if featured_posts and posts.number == 1 and 'category' not in request.GET %}
+                    {% comment %}
+                        An even number of posts, n, is required "in between" featured posts.
+                        This is dependent on the number of featured posts.
+                        We therefore display a featured post when `forloop.counter` is odd.
+                        With a pagination of 10:
```
zerolab, line 18: suggestion: simplify the case comments so they are clear for non-developers.

I may be thick, but I find it difficult to follow the logic.

As far as I can see the only special case is when there are more featured posts than regular blog posts on the first page, which we should never reach at this point

engineervix: Thanks! Done in 5519e0f

wagtailio/project_styleguide/templates/patterns/pages/blog/blog_index_page.html
```diff
@@ -9,8 +9,55 @@
...
+                            {% include 'patterns/pages/blog/includes/blog_post.html' with post=featured_posts.2 featured=True %}
+                        {% elif forloop.counter == 5 and featured_posts|length == 2 or forloop.counter == 5 and featured_posts|length == 3 %}
+                            {% include 'patterns/pages/blog/includes/blog_post.html' with post=featured_posts.1 featured=True %}
+                        {% elif forloop.counter == 7 and featured_posts|length >= 4 %}
+                            {% include 'patterns/pages/blog/includes/blog_post.html' with post=featured_posts.3 featured=True %}
+                        {% elif forloop.counter == 9 and featured_posts|length == 5 %}
+                            {% include 'patterns/pages/blog/includes/blog_post.html' with post=featured_posts.4 featured=True %}
+                        {% elif forloop.counter == 9 and featured_posts|length == 3 %}
+                            {% include 'patterns/pages/blog/includes/blog_post.html' with post=featured_posts.2 featured=True %}
+                        {% endif %}
```
zerolab, line 58: How about something like:

```django+html
{% with num_featured_posts=featured_posts|length %}
    {% if forloop.first %}
        {% include 'patterns/pages/blog/includes/blog_post.html' with post=featured_posts.0 featured=True %}
    {% elif forloop.counter == 3 and num_featured_posts >= 4 %}
        {% include 'patterns/pages/blog/includes/blog_post.html' with post=featured_posts.1 featured=True %}
    {% elif forloop.counter == 5 %}
        {% if num_featured_posts >= 4 %}
            {% include 'patterns/pages/blog/includes/blog_post.html' with post=featured_posts.2 featured=True %}
        {% elif num_featured_posts == 2 or num_featured_posts == 3 %}
            {% include 'patterns/pages/blog/includes/blog_post.html' with post=featured_posts.1 featured=True %}
        {% endif %}
    {% elif forloop.counter == 7 and num_featured_posts >= 4 %}
        {% include 'patterns/pages/blog/includes/blog_post.html' with post=featured_posts.3 featured=True %}
    {% elif forloop.counter == 9 %} 
        {% if num_featured_posts == 5 %}
            {% include 'patterns/pages/blog/includes/blog_post.html' with post=featured_posts.4 featured=True %}
        {% elif num_featured_posts == 3 %}
            {% include 'patterns/pages/blog/includes/blog_post.html' with post=featured_posts.2 featured=True %}
        {% endif %}
    {% endif %}
{% endwith %}
```

Note: not tested locally.


Other than that, let's merge it/

engineervix: Thanks, tested locally and working well!

PR #237, opened 2022-08-22
Error pages

chris-lawton:
Styles 404 and 500 available to view locally [http://localhost:8000/test404/](http://localhost:8000/test404/) and [http://localhost:8000/test500/](http://localhost:8000/test500/)


wagtailio/static/sass/components/_error-404.scss
```diff
@@ -0,0 +1,48 @@
+.error-404 {
+    position: relative;
+
+    &__intro,
+    &__heading {
```
William-Blackie, line 5: Question: Is it better to use two classes that do the same thing in this case or one?

William-Blackie (changes requested):

wagtailio/project_styleguide/templates/patterns/500.html
```diff
@@ -1,18 +1,102 @@
...
+                <h1>Something went wrong</h1>
+                <p>We&rsquo;re sorry! Our server is acting up right now.</p>
```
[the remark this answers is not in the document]
chris-lawton: There isn't one for the 500 template only the 404. Not sure a 500 needs one but we can run this past Ben when he's back.

wagtailio/static/sass/components/_error-404.scss
chris-lawton: `__intro` requires [some additional styles](https://github.com/wagtail/wagtail.org/blob/50-error-pages/wagtailio/static/sass/components/_error-404.scss#L17) which is why they're different. I also like that `__heading` clearly refers to a `<h1>` tag and intro a `<p>`.

 I guess you could have a new class for grid alignment, which thinking about it we have a ticket for in the backlog for - to have grid helper classes as at the moment we have some repetition of aligning items using `grid-column: x / span x`

wagtailio/urls.py
```diff
@@ -58,6 +58,7 @@
 ):
     urlpatterns += [
         path("pattern-library/", include("pattern_library.urls")),
+        path("test500", test_500_error),
```
William-Blackie, line 61: Hey @zerolab, 

how do you feel about a testing view for the 500 error page?

zerolab: 👍 for the approach

PR #238, opened 2022-08-23
Mobile search 

chris-lawton:
Implements the mobile search design. Can be viewed on any page outside the PL on a mobile view.


wagtailio/project_styleguide/templates/patterns/includes/search-modal.html
```diff
@@ -0,0 +1,15 @@
+<div class="modal modal--search micromodal-slide" id="site-search-modal" aria-hidden="true">
+    <div class="modal__overlay" tabindex="-1" data-micromodal-close>
+        <div class="modal__container" role="dialog" aria-modal="true">
+            <main class="modal__content">
+                <button class="modal__close-button" aria-label="Close the Search" data-micromodal-close>
+                    <svg class="modal__close-icon" aria-hidden="true">
```
William-Blackie, line 6: Note: doesn't this icon have a purpose that is not just functional? I thought this would not be hidden

wagtailio/project_styleguide/templates/patterns/includes/header.html
```diff
@@ -13,12 +13,24 @@
...
             </svg>
         </button>
 
+        {# Mobile Search #}
+        <button class="header__button header__button--search u-mobile-only" aria-expanded="false" aria-controls="mobile-search-form" aria-label="Open the search" data-site-search-mobile>
+            <svg class="header__search-icon header__search-icon--open" aria-hidden="true">
+                <use xlink:href="#magnifying-glass"></use>
+            </svg>
+
+            <svg class="header__search-icon header__search-icon--close" aria-hidden="true">
```
William-Blackie, line 29: Note: I think aria-hidden might not be the correct usage here?

wagtailio/project_styleguide/templates/patterns/includes/header.html
```diff
@@ -13,12 +13,24 @@
 
 
     <div class="header__col-two">
-        <button class="header__button header__button--search" aria-label="Open the search" data-site-search>
+        {# Desktop Search #}
```
William-Blackie, line 16: Note: on mobile when on search_results.html, I think this mobile view breaks the normal use case as the search results block is already shown, what do you think?

wagtailio/project_styleguide/templates/patterns/includes/search-modal.html
chris-lawton: It has a purpose for users not using assistive tech but for users using it then it doesn't serve a purpose - this is where the aria-label on the parent button will be announced instead.

wagtailio/project_styleguide/templates/patterns/includes/header.html
chris-lawton: Similar to the above - the aria-label on the parent button will be announced so there's no need to have this icon in the accessibility tree.

wagtailio/project_styleguide/templates/patterns/includes/header.html
chris-lawton: I see what you mean - there's a double search of sorts. I don't think it's too much of an issue having thought about it. If a user on mobile landed on the search results page then they'd probably use the search input that on the page template but I don't think it's an issue if they opened the other search and used that instead.

PR #244, opened 2022-08-25
Text & Media CTA

chris-lawton:
`cta_block.html` contains a `grid` class for when it's used as a SF but in this instance it's causing alignment issues when used a child of a SF. Using `button.html` resolves this.

Before: 

Screenshot 2022-08-25 at 10 31 30

After:

Screenshot 2022-08-25 at 10 31 19


zerolab:
@William-Blackie can you add this to your list?

nicklee:
I've actioned this seperately as part of the design review

PR #246, opened 2022-08-26
Menus

engineervix:
This PR wires up the wagtail admin interface to allow content editors to specify content for the updated Main and Footer menus

- [X] add new **navigation** app
- [X] add a `FooterMenuSnippet`
- [x] add a `MainMenuItemSnippet`
- [x] add a `MainMenuSnippet`, which allows one to select Main Menu Items 
- [x] add `NavigationSettings` with snippet choosers for
  - [x] the main nav, and 
  - [X] the footer nav.
  - [X] the get started menu

<details><summary>Screenshots</summary>

Untitled

image

image

image

image

image

</details>

### UPDATE 2022-09-28

The the main navigation structure has been revised (see 1f6ef38f, and https://github.com/wagtail/wagtail.org/pull/246#discussion_r960893625). Check out the GIFs below:

<details><summary>GIF | Wagtail Admin</summary>

main_menu_nav__001 min

</details>

<details><summary>GIF | Template</summary>

**Note**: this is to be updated in a separate MR that focuses on the frontend

main_menu_nav__002 min

</details>


wagtailio/navigation/templatetags/navigation_tags.py
```diff
@@ -0,0 +1,14 @@
+from django import template
+
+register = template.Library()
+
+
+@register.inclusion_tag("patterns/includes/menu_footer.html", takes_context=True)
+def footer_navigation(context):
+    request = context["request"]
+    return {
+        "footer_navigation": context["settings"]["navigation"][
```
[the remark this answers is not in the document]
engineervix: Perhaps this could be addressed by following the approach used here?
 https://github.com/wagtail/wagtail.org/blob/20db37b74dd38a9e9082277289534b6909275a56/wagtailio/project_styleguide/pattern_contexts.py#L8-L10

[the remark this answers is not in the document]
zerolab: I am not fully enamored by that approach as it makes two queries.

that bit can be 
```python
if nav := = NavigationSettings.objects.first():
  context["settings"] = {"navigation": {"NavigationSettings": nav}} 
```

note `:=` is the [walrus operator](https://realpython.com/python-walrus-operator/)

---

For this, it is best to wrap it in a try/except:

```python
try:
   footer_navigation = context["settings"]["navigation"]["NavigationSettings"].footer_navigation
except (KeyError, AttributeError):
   footer_navigation = None

return {
   "footer_navigation": footer_navigation,
   "request": request
}
```

question: do you need the request in the template context?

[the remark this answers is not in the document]
engineervix: > question: do you need the request in the template context?

You're right, I don't think it's needed! I've removed it in 4b347d68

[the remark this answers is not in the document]
engineervix: > For this, it is best to wrap it in a try/except:

Thanks, done in 4b347d6

wagtailio/navigation/templatetags/navigation_tags.py
```diff
@@ -0,0 +1,53 @@
...
+def main_navigation(context):
+    try:
+        main_navigation = context["settings"]["navigation"][
+            "NavigationSettings"
+        ].main_navigation
+    except (KeyError, AttributeError):
+        return {}
+
+    menus = []
+    for item in main_navigation.menu_items.all():
```
zerolab, line 30: If you do not have a main nav defined, this fails with `'NoneType' object has no attribute 'menu_items'`

wagtailio/project_styleguide/templates/patterns/includes/menu_main.html
```diff
@@ -0,0 +1,38 @@
...
+                                <h3>{{ cta.text }}</h3>
+                                <p>{{ cta.sub_heading }}</p>
+                                {{ cta.description }}
+                            </a>
+                        {% endwith %}
+                    {% endfor %}
+                </div>
+            </li>
+        {% empty %}
+            No navigation links
```
zerolab, line 35: nitpick: this should not be output

wagtailio/project_styleguide/templates/patterns/includes/menu_main.html
```diff
@@ -0,0 +1,38 @@
...
+<nav class="">
+    <ul class="">
+        {% for menu_item in menus %}
+            <li class="">
+                <span>{{ menu_item.name }}</span>
+                <ul class="">
+                    {% for item in menu_item.nav_items %}
+                        {% with item.value as nav_item %}
+                            <li class="">
+                                <a href="{{ nav_item.url }}" title="{{ nav_item.text }}">
```
zerolab, line 12: This explodes because the main nav does not validate that a page or url is added to each nav item

wagtailio/navigation/blocks.py
```diff
@@ -0,0 +1,53 @@
...
+
+# --------------- Footer Navigation ---------------
+
+
+class NavLinkBlock(CTALinkMixin):
+    text = blocks.CharBlock(max_length=100)
+    cta_page = blocks.PageChooserBlock(label="Page", required=False)
+    cta_url = blocks.URLBlock(label="URL", required=False)
+
+    class Meta:
```
zerolab, line 14: There is no validation to ensure either page or url is added

wagtailio/navigation/models.py
```diff
@@ -0,0 +1,126 @@
...
+
+    def __str__(self):
+        return self.name
+
+    class Meta:
+        verbose_name = "Footer menu"
+
+
+@register_snippet
+class MainMenuItemSnippet(models.Model):
```
zerolab, line 41: nitpick: let's call it just `MainMenuItem`, ditto for `MainMenuSnippet` and `FooterMenuSnippet`

zerolab: as an aside.. perhaps `MainMenuSection` is more appropriate, but this is not a blocker

wagtailio/navigation/templatetags/navigation_tags.py
engineervix: Thanks for this. Fixed in 5d11060

wagtailio/project_styleguide/templates/patterns/includes/menu_main.html
engineervix: Thanks, removed in d894889

wagtailio/project_styleguide/templates/patterns/includes/menu_main.html
engineervix: Thanks! URL Validation fixed in 24d3f6c

wagtailio/navigation/models.py
engineervix: Thanks, done some bit of refactoring in 080d08f

wagtailio/project_styleguide/pattern_contexts.py
```diff
@@ -0,0 +1,9 @@
+from pattern_library import register_context_modifier
+
+from wagtailio.navigation.models import NavigationSettings
+
+
+@register_context_modifier
+def add_navigation(context, request):
+    if nav := NavigationSettings.objects.first():
```
William-Blackie, line 8: Note: this is not a full review. I see what you're trying to do here in terms of reducing queries but this will result in an error if a NavigationSettings instance is not found

wagtailio/navigation/apps.py
```diff
@@ -0,0 +1,6 @@
+from django.apps import AppConfig
+
+
+class NavigationConfig(AppConfig):
+    name = "wagtailio.navigation"
+    default_auto_field = "django.db.models.AutoField"
```
William-Blackie, line 6: Suggestion: Is there a reason not to set this globally? 

`DEFAULT_AUTO_FIELD = 'django.db.models.AutoField'` Add this to `base.py`

wagtailio/project_styleguide/templates/patterns/base_page.html
```diff
@@ -1,5 +1,5 @@
 {% extends "patterns/base.html" %}
-{% load static wagtailcore_tags wagtailimages_tags wagtail_ab_testing_tags wagtailio_utils %}
+{% load static wagtailcore_tags wagtailimages_tags wagtail_ab_testing_tags wagtailio_utils navigation_tags %}
```
William-Blackie, line 2: Question: What are we adding this template tag for here? I don't think there's a template tag below that isn't in it's own template.

wagtailio/project_styleguide/templates/patterns/includes/menu_main.html
```diff
@@ -0,0 +1,36 @@
...
+                        {% with item.value as cta %}
+                            <a href="{{ cta.url }}" title="{{ cta.text }}">
+                                {% include "patterns/components/icon/icon.html" with icon=cta.icon classes="icon-link__svg" %}
+                                <h3>{{ cta.text }}</h3>
+                                <p>{{ cta.sub_heading }}</p>
+                                {{ cta.description }}
+                            </a>
+                        {% endwith %}
+                    {% endfor %}
+                </div>
```
William-Blackie, line 32: Suggestion: Great Job here Victor, it would be great if you could include the main_menu.yaml file with the markup for this FE.

wagtailio/navigation/templatetags/navigation_tags.py
```diff
@@ -0,0 +1,54 @@
...
+@register.inclusion_tag("patterns/includes/menu_main.html", takes_context=True)
+def main_navigation(context):
+    menus = []
+
+    try:
+        main_navigation = context["settings"]["navigation"][
+            "NavigationSettings"
+        ].main_navigation
+
+        for item in main_navigation.menu_sections.all():
```
William-Blackie, line 29: Suggestion: You could replace this for-loop with extend instead; 

```
menus.extend({"name": item.menu_section.name, "nav_items": item.menu_section.nav_items, "call_to_action": item.menu_section.call_to_action,} for item in main_navigation.menu_sections.all()
```

wagtailio/navigation/models.py
```diff
@@ -0,0 +1,126 @@
...
+    ]
+
+    def __str__(self):
+        return self.name
+
+    class Meta:
+        verbose_name = "Footer menu"
+
+
+@register_snippet
```
William-Blackie, line 40: @zerolab I'm happy to flesh this out with you, personally, I don't think navigation UI is that well suited here for snippets. It works but I think the workflow is awkward.

To set this we'd need to in this order; create Footer, Main menu and Get started items and then add them to the main add them to navigation settings. I think that it might be more appropriate to do something more akin to a menu settings hook using the clusterable model (so we can create different menus for the sites? not sure if that's needed) and then adding the streamfields there in one place.

What do you think?

zerolab: I am not very keen on one massive Settings instance with large StreamFields.
The idea behind using snippets is for content staging and for previews, should we need to.

`MainMenu` could potentially be simplified to be one big StreamField, though, as I agree the workflow can be improved. Basically `MainMenuSection` becoming a StreamField block in `MainMenu`, rather than FK

wagtailio/navigation/templatetags/navigation_tags.py
zerolab: reformatted:

```python
menus.extend(
    {
        "name": item.menu_section.name,
        "nav_items": item.menu_section.nav_items,
        "call_to_action": item.menu_section.call_to_action,
    }
    for item in main_navigation.menu_sections.all()
)
```

which is very close the for-loop and I don't see any gains in readability or other? 🤔

wagtailio/project_styleguide/templates/patterns/base_page.html
engineervix: Good point, removed in 25cd9d15

wagtailio/navigation/apps.py
engineervix: Done in 81adc069

wagtailio/project_styleguide/pattern_contexts.py
engineervix: By using `.first()`, we avoid the error, since, `None` will  be returned if there's no instance.

wagtailio/navigation/templatetags/navigation_tags.py
engineervix: I did my own test:

```python
import timeit
from faker import Faker

fake = Faker()

menu_sections = [
    {
        "name": fake.name(),
        "call_to_action": fake.sentence(),
        "nav_items": fake.words(),
    }
    for _ in range(40)
]


def append_fn():
    menus = []
    for item in menu_sections:
        menus.append(item)


def extend_fn():
    menus = []
    menus.extend(item for item in menu_sections)


print(min(timeit.repeat(lambda: append_fn())))
print(min(timeit.repeat(lambda: extend_fn())))
```

and `append` wins, although the difference is small.

```console
❯ python append_vs_extend.py
2.6496993280015886
2.758143986982759
❯ python append_vs_extend.py
2.60596901900135
2.7336814689915627
```

wagtailio/navigation/models.py
engineervix: I've updated this one in 1f6ef38f

wagtailio/project_styleguide/templates/patterns/includes/menu_main.html
engineervix: Thanks, I've added it in efcde89d

Though I think it won't work in the pattern library, at least for now, because we are actually using data from a NavigationSettings instance:

https://github.com/wagtail/wagtail.org/blob/d6d778602ac242465307bfc8085d55e51d0ff031/wagtailio/project_styleguide/pattern_contexts.py#L6-L9

wagtailio/navigation/models.py
```diff
@@ -0,0 +1,80 @@
...
+class MainMenu(ClusterableModel):
```
bcdickinson, line 33: Now that we're not using snippets for the sections, this could just be a `Model` again. Not a dealbreaker, but simpler is betterer :smile:

wagtailio/project_styleguide/pattern_contexts.py
bcdickinson: Do we want to be pulling in the real data for use in the pattern library? It seems like we'd be better off hard-coding something representative here and not touching the DB at all. This isn't a show-stopper for now, though.

bcdickinson (approved):
Thanks, this is looking good now. I've got a couple of minor nitpicks, but this is good enough to merge as is.

wagtailio/navigation/models.py
zerolab: `MainMenu` is still a snippet

zerolab (approved):

PR #248, opened 2022-08-26
Example Fix menus

William-Blackie:
Do not merge, this is an example of overriding the pattern library templates.


wagtailio/navigation/templatetags/navigation_tags.py
```diff
@@ -4,7 +4,7 @@
 
 
 @register.inclusion_tag("patterns/includes/menu_footer.html", takes_context=True)
-def footer_navigation(context):
+def footer_navigation(context, block):
```
William-Blackie, line 7: This change here allows us to map a value over in the pattern library. We could also achieve this with context overrides, up to you what we do really.

zerolab:
@William-Blackie is this still needed?

PR #254, opened 2022-08-31
Fix/footer sign up form styling

William-Blackie:


wagtailio/static/sass/components/_sign-up-form.scss
```diff
@@ -71,6 +71,10 @@
         &::placeholder {
             color: $color--off-black;
         }
+
+        &:focus-visible {
+            outline: none;
+        }
```
William-Blackie, line 77: This removes the yellow outline, not specifically requested but was missed in review.

zerolab: isn't that an accessibility no-no?

William-Blackie: You're right, https://www.w3.org/WAI/WCAG22/Understanding/focus-visible.html. I should have known this one!

Issue #260, opened 2022-09-12
Odd sticky header behavior on mobile with a tall sitewide alert


joshuadavidthomas:
When visiting https://wagtail.org on a mobile device, if there is a tall sidewide alert the header behaves in odd ways. 

If the mobile menu is closed and you scroll down with enough speed to trigger hiding the header, it shifts up to hide, but because of the height of the sitewide alert it's still visible, just double the height instead.

If the mobile menu is open, the header appears not at the top of the screen but towards the top obscuring some of the menu items.

## Video

https://user-images.githubusercontent.com/19896267/189369193-dbcc5ab5-5e0c-4aa4-8cb3-8699c461ace0.mp4

thibaudcolas:
@joshuadavidthomas thanks for the report! We’re working on a redesign of the site due to launch next week where this will no longer be an issue, so I’ll close this now.

PR #261, opened 2022-09-12
add fixed position to header when mobile menu is open

joshuadavidthomas:
Fixes the second issue of #260 and makes the header fixed at the top of the screen when the mobile menu is open.


zerolab:
@nicklee one to consider when working on nav

thibaudcolas:
@joshuadavidthomas I can confirm this has been superseded by our current redesign so I’ll close this now.

PR #264, opened 2022-09-28
Bugfix for broken wagtail admin page previews

engineervix:
The Preview function is not working because it is asking for credentials and will not accept the default password

### DIFF
--- a/wagtailio/settings/base.py
+++ b/wagtailio/settings/base.py
@@ -149,8 +149,8 @@ MEDIA_URL = env.get("MEDIA_URL", "/media/")
 # Basic auth settings
 if os.environ.get("BASIC_AUTH_ENABLED", "false").lower() == "true":
     MIDDLEWARE.insert(0, "baipw.middleware.BasicAuthIPWhitelistMiddleware")
-    BASIC_AUTH_LOGIN = env.get("BASIC_AUTH_LOGIN")
-    BASIC_AUTH_PASSWORD = env.get("BASIC_AUTH_PASSWORD")
+    BASIC_AUTH_LOGIN = env.get("BASIC_AUTH_LOGIN", "wagtailorg")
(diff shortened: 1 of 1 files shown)


wagtailio/settings/base.py
```diff
@@ -151,6 +151,10 @@
     MIDDLEWARE.insert(0, "baipw.middleware.BasicAuthIPWhitelistMiddleware")
     BASIC_AUTH_LOGIN = "wagtailio"
     BASIC_AUTH_PASSWORD = "showmewagtailio"
```
zerolab, line 153: while here, it would be great to update this bit to properly get the credentials from the environment variables, rather than hardcoded values

engineervix: Thanks @zerolab, updated in 6ae8d4a3

wagtailio/settings/base.py
```diff
@@ -149,8 +149,8 @@
 # Basic auth settings
 if os.environ.get("BASIC_AUTH_ENABLED", "false").lower() == "true":
     MIDDLEWARE.insert(0, "baipw.middleware.BasicAuthIPWhitelistMiddleware")
-    BASIC_AUTH_LOGIN = "wagtailio"
-    BASIC_AUTH_PASSWORD = "showmewagtailio"
+    BASIC_AUTH_LOGIN = env.get("BASIC_AUTH_LOGIN")
+    BASIC_AUTH_PASSWORD = env.get("BASIC_AUTH_PASSWORD")
```
zerolab, line 153:
```suggestion
    BASIC_AUTH_LOGIN = env.get("BASIC_AUTH_LOGIN", "wagtailorg")
    BASIC_AUTH_PASSWORD = env.get("BASIC_AUTH_PASSWORD", "showmewagtailorg")
```
it is good practice to have a sensible fallback

PR #265, opened 2022-09-29
Wagtail 4.0 upgrade

bcdickinson:
This PR upgrades the project and it's dependencies (as far as possible) to be compatible with Wagtail 4.0.

To make life easier for myself in terms of being able to easily detect incompatible versions and improve build reproducability, I migrated from `pip`/requirements.txt to `poetry==1.2.1` for dependency management first.

I hit the following snags while upgrading the various `wagtail*` 3rd party packages:

- `wagtail-ab-testing` explicitly does not support `wagtail>=3` so `poetry` refuses to install it. The are [open](https://github.com/torchbox/wagtail-ab-testing/pull/48) [PRs](https://github.com/torchbox/wagtail-ab-testing/pull/49) to add support (hat tip to @nickmoreton), but they didn't make it in in time for this PR, so for the time being I've just removed A/B testing from this project.
- `wagtail-airtable` has some incompatibiliites with Wagtail 4.0 so the sync currently doesn't work and **should not** be used. I've elected not to remove it at this point to avoid losing the `airtable_record_id` data on `AirtableMixin` subclasses.
- `wagtail-content-import`, `wagtail-image-import`, and `wagtailaltgenerator` all superficially appear to be compatible with Wagtail 4.0 with the caveat that I haven't done extensive manual testing and the packages themselves don't officially support or test against it.
- `wagtail-transfer` doesn't officially support or explicitly test against Wagtail 4.0, but the nightly builds against `main` are currently passing, so we're on pretty safe ground here
- `wagtailmedia` is fine, it explicitly supports and tests against Wagtail 4.0

I've raised tickets to reinstate `wagtail-ab-testing` following it being brought up to date and to add 4.0 support to `wagtail-airtable`.


Dockerfile
```diff
@@ -9,15 +9,19 @@ RUN npm ci --no-optional --no-audit --progress=false
...
+# Build Python app - this stage is a common base for the prod and dev stages
 FROM python:3.8-bullseye AS backend
 
-ENV PYTHONUNBUFFERED=1 \
-    PYTHONPATH=/app \
-    DJANGO_SETTINGS_MODULE=wagtailio.settings.production \
+ARG POETRY_VERSION=1.2.1
+
+ENV DJANGO_SETTINGS_MODULE=wagtailio.settings.production \
+    GUNICORN_CMD_ARGS="--max-requests 1200 --access-logfile -" \
```
zerolab, line 19: would be nice to add the jitter param as well (https://adamj.eu/tech/2019/09/19/working-around-memory-leaks-in-your-django-app/#gunicorn) while at it

bcdickinson: Good idea, done.

wagtailio/core/wagtail_hooks.py
```diff
@@ -2,37 +2,16 @@
...
-        """
-    )
-
-
-hooks.register("insert_editor_js", editor_js)
-
-
 def editor_css():
     return format_html(
         '<link rel="stylesheet" href="'
```
zerolab, line 17: might be nice to replace the raw font-awesome usage with https://github.com/allcaps/wagtail-font-awesome-svg

pyproject.toml
```diff
@@ -0,0 +1,57 @@
...
+license = "BSD-3-Clause"
+readme = "README.md"
+packages = [{include = "wagtailio"}]
+
+[build-system]
+requires = ["poetry-core"]
+build-backend = "poetry.core.masonry.api"
+
+[tool.poetry.dependencies]
+python = "^3.8"
```
zerolab, line 15: question: any reason not to make the leap to 3.9 if not 3.10?

docker-compose.yml
```diff
@@ -12,7 +12,7 @@ services:
       SECRET_KEY: fake
       DJANGO_SETTINGS_MODULE: wagtailio.settings.dev
       DATABASE_URL: postgres://wagtailio:wagtailio@db:5432/wagtailio
-
+      PRIMARY_HOST: http://localhost:8000
```
zerolab, line 15: note: #268 changes the healthcheck to use `pg_isready -U wagtailio` to avoid noise in the logs

Makefile
```diff
@@ -1,36 +1,44 @@
-.PHONY: setup rebuild start superuser migrations migrate pull-production-data pull-production-media pull-staging-data pull-staging-media
-
-setup:
-	make rebuild
+.PHONY: setup
+setup: rebuild
```
zerolab, line 2: my preference would be to use something like https://github.com/wagtail/wagtail-localize/blob/main/Makefile whereby running `make` lists the available commands so you can get the list of available commands


aside: is it time to switch to using `docker compose` instead of `docker-compose` ?

zerolab (commented):
The vast majority of changes make sense 👍

Left a number of comments/questions

pyproject.toml
bcdickinson: No, done.

Makefile
bcdickinson: That'd be nice, but perhaps outside the scope here. I'd prefer to leave it as `docker-compose` for now. This isn't making use of any newer features that necessitate plugin rather than the standalone and people on the newer version already can use `compose-switch` or similar: https://github.com/docker/compose-switch

Makefile
```diff
@@ -1,36 +1,44 @@
...
 	docker-compose run web django-admin migrate
 	docker-compose run web django-admin createcachetable
 
+.PHONY: rebuild
 rebuild:
-	bash -c "docker-compose build --build-arg UID=$$(id -u) --build-arg DOCKER_GID=$$(ls -ln /var/run/docker.sock | awk '{print $$4}')"
+	bash -c "docker-compose build --build-arg UID=$$(id -u) --build-arg GID=$$(id -g)"
```
zerolab, line 8:
```
 => ERROR [backend 5/6] RUN groupadd --gid 20 wagtailio     && useradd --create-home --uid 503 --gid 20 wagtailio                                                   0.2s
------
 > [backend 5/6] RUN groupadd --gid 20 wagtailio     && useradd --create-home --uid 503 --gid 20 wagtailio:
#0 0.201 groupadd: GID '20' already exists
------
failed to solve: executor failed running [/bin/sh -c groupadd --gid $GID wagtailio     && useradd --create-home --uid $UID --gid $GID wagtailio]: exit code: 4
make: *** [rebuild] Error 17
```

pyproject.toml
bcdickinson: _record scratch_ Actually, the ancient version of `fab` that we're using breaks on Python>=3.9 due to `collections.abc.Mapping` not being importable from `collections` any more, so I've reverted that change and we're still on 3.8 for now.

pyproject.toml
```diff
@@ -36,12 +36,12 @@ requests = "~2.28.1"
 scout-apm = "~2.26.1"
 sentry-sdk = "~1.5.12"
 wagtail = "~4.0.2"
-wagtail-airtable = "~0.2.1"
-wagtail-content-import = "~0.6.0"
+wagtail-airtable = "~0.2.1"  # This version isn't actually compatible with Wagtail 4 - https://torchbox.monday.com/boards/1123315381/pulses/1130533301
```
zerolab, line 39: Best to drop itdue to the notes you added.

suggestion: let's avoid Torchbox-specific links that others cannot access. Those could go in a wiki page on the repo

zerolab: doing some 🕵️‍♂️ work and will get back to this

Makefile
bcdickinson: Should be fixed by https://github.com/wagtail/wagtail.org/pull/265/commits/c9d247fd2f59bab0706cadb6c133612a54b77bba

pyproject.toml
zerolab: discussed and these all should be removed: `wagtail-airtable`, `wagtail-content-import`, `wagtail-image-import`, `wagtail-transfer`.

`wagtailaltgenerator` seems useful so we could leave it

PR #268, opened 2022-10-03
Docker compose update

zerolab:
fixes the healthcheck command, and changes the use of `wagtailio` to `wagtailorg` where possible


bcdickinson (approved):
Looks good to me.

PR #73, opened 2020-10-15
New services template [FE]

janehughes:
image

I couldn't see a way of building the new template without having the BE in place first (as there's no pattern lib), so I've created the new services template alongside the standard page template and added an if statement based on the page title for now.

PR #272, opened 2022-10-04
Fix usage of optional CTA block

engineervix:
The main purpose of this Pull Request is to provide a fix for optional CTA Blocks.

- [`CTABlock`](https://github.com/wagtail/wagtail.org/blob/b4d3483d3e59dd93a5097082e15121720e81dc05/wagtailio/core/blocks.py#L243-L255) is **required** by default, because its [`required`](https://github.com/wagtail/wagtail.org/blob/b4d3483d3e59dd93a5097082e15121720e81dc05/wagtailio/core/blocks.py#L249-L250) property is set to `True`
- Therefore, attempting to use it in the manner [`cta = CTABlock(required=False`](https://github.com/wagtail/wagtail.org/blob/b4d3483d3e59dd93a5097082e15121720e81dc05/wagtailio/core/blocks.py#L263), as has been the case prior to this PR, doesn't work
- This PR thus introduces an `OptionalCTABlock` which extends `CTABlock`, but overrides the `required` property, setting it to `False`.
- Now, if an optional CTA is required, we simply use the new `OptionalCTABlock`
- This change inevitably results in a migration for the ContentPage model, because of changes in the `body` field

This change affects the following blocks:

1. CardBlock
2. HeadlineBlock
3. HighlightBlock
4. IconBulletBlock
5. TextAndMediaBlock

In addition

- The `LogoCardBlock` is fixed by renaming the `heading` field to `text`, while still labelling it as _Heading_. This is because `LogoCardBlock` extends `CTALinkMixin`, which expects a `text` field in the StructBlock definition. The block's template is also updated to reflect this change.
- There are a couple of bug fixes made on the following templates
  - `patterns/components/streamfields/cards/logo_card_block.html` -- pass missing `url` context to the included button block template
  - `patterns/components/streamfields/text_and_media_block/text_and_media_block.html` -- update the condition for rendering the CTA, and include the correct template (See attached screenshots that necessitated this change)
- Some documentation changes are made to the template `patterns/components/streamfields/cards/related_page_card_block.html` to reflect changes made to the LogoCardBlock and its template.

<details><summary>Screenshots: text_and_media_block fix</summary>

Before

text_and_media_block_cta_before

After

text_and_media_block_cta_after

</details>

**Tickets reference**

- [#72](https://torchbox.monday.com/boards/1123315381/pulses/1129515803)
- [#73](https://torchbox.monday.com/boards/1123315381/pulses/1129514718)


bcdickinson (approved):
Great work, thanks Victor!

PR #285, opened 2022-10-14
Template tidy-ups

zerolab:
This PR:

* fixes `base_page.html` template inheritance. now we get all the metadata glory
* tidies up various tracker usage
* re-adds missing static images (clipboard svg and default social image)
* optimises static images
* re-adds the cookie message
* re-adds the sitewide alert functionality

To-Do:
* [x] navigation template fragment caching - #286

PR #286, opened 2022-10-14
Nav fragment caching and renditions cache

zerolab:
Builds on #285 and adds template fragment cache for main and footer nav.

The cache is cleared when the navigation settings is saved. It is also cleared when one saves the main or footer nav snippet that is in use.

Note: using snippets for main/footer nav so as to implement previews now that we are on Wagtail 4

### DIFF
--- a/wagtailio/navigation/models.py
+++ b/wagtailio/navigation/models.py
@@ -50,7 +50,6 @@ class MainMenu(ClusterableModel):
         super().save(**kwargs)
 
         if NavigationSettings.objects.filter(main_navigation=self).exists():
-            print("delete")
             cache.delete(make_template_fragment_key("primarynav"))
 
     def __str__(self):
--- a/wagtailio/navigation/models.py
+++ b/wagtailio/navigation/models.py
@@ -1,3 +1,5 @@
+from django.core.cache import cache
+from django.core.cache.utils import make_template_fragment_key
 from django.db import models
 
 from modelcluster.models import ClusterableModel
@@ -19,6 +21,12 @@ class FooterMenu(models.Model):
         FieldPanel("sections"),
     ]
 
+    def save(self, **kwargs):
+        super().save(**kwargs)
+
+        if NavigationSettings.objects.filter(main_navigation=self).exists():
+            cache.delete(make_template_fragment_key("footernav"))
+
     def __str__(self):
         return self.name
 
@@ -28,7 +36,6 @@ class FooterMenu(models.Model):
 
 @register_snippet
 class MainMenu(ClusterableModel):
-
     name = models.CharField(max_length=255)
     menu_sections = StreamField(
         [("menu_section", MainMenuSectionBlock())], use_json_field=True
@@ -38,6 +45,13 @@ class MainMenu(ClusterableModel):
         FieldPanel("name"),
         FieldPanel("menu_sections", classname="collapsible"),
     ]
+
+    def save(self, **kwargs):
+        super().save(**kwargs)
+
+        if NavigationSettings.objects.filter(main_navigation=self).exists():
+            print("delete")
+            cache.delete(make_template_fragment_key("primarynav"))
 
     def __str__(self):
         return self.name
@@ -75,3 +89,12 @@ class NavigationSettings(BaseSiteSetting, ClusterableModel):
         FieldPanel("main_navigation"),
         FieldPanel("footer_navigation"),
     ]
+
+    def save(self, **kwargs):
+        super().save(**kwargs)
+
+        keys = [
+            make_template_fragment_key(key, vary_on=[self.site.pk])
+            for key in ["primarynav", "footernav"]
+        ]
+        cache.delete_many(keys)
--- a/wagtailio/navigation/models.py
+++ b/wagtailio/navigation/models.py
@@ -25,7 +25,11 @@ class FooterMenu(models.Model):
         super().save(**kwargs)
 
         if NavigationSettings.objects.filter(main_navigation=self).exists():
-            cache.delete(make_template_fragment_key("footernav"))
+            cache.delete(
+                make_template_fragment_key(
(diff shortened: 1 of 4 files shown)


wagtailio/navigation/models.py
```diff
@@ -39,6 +46,13 @@ class MainMenu(ClusterableModel):
         FieldPanel("menu_sections", classname="collapsible"),
     ]
 
+    def save(self, **kwargs):
+        super().save(**kwargs)
+
+        if NavigationSettings.objects.filter(main_navigation=self).exists():
+            print("delete")
```
RealOrangeOne, line 53: Issue: Print

wagtailio/settings/base.py
```diff
@@ -264,12 +264,36 @@
...
+    connection_pool_kwargs = {}
+
+    if REDIS_URL.startswith("rediss"):
+        # Heroku Redis uses self-signed certificates for secure redis conections.
+        # https://stackoverflow.com/a/66286068
+        # When using TLS, we need to disable certificate validation checks.
+        connection_pool_kwargs["ssl_cert_reqs"] = None
+
+    redis_options = {
+        "IGNORE_EXCEPTIONS": True,
```
RealOrangeOne, line 280: Because of this, we should add `DJANGO_REDIS_LOG_IGNORED_EXCEPTIONS = True`

https://git.torchbox.com/internal/wagtail-kit/-/blob/master/wagtailkit_repo_name/wagtailkit_repo_name/settings/base.py#L200

wagtailio/navigation/models.py
```diff
@@ -19,6 +21,12 @@ class FooterMenu(models.Model):
         FieldPanel("sections"),
     ]
 
+    def save(self, **kwargs):
```
RealOrangeOne, line 24: Suggestion: Perhaps this would be better handled using a signal or hook than custom `save` method?

wagtailio/navigation/models.py
```diff
@@ -19,6 +21,12 @@ class FooterMenu(models.Model):
...
+        super().save(**kwargs)
+
+        if NavigationSettings.objects.filter(main_navigation=self).exists():
```
RealOrangeOne, line 27: Suggestion: Perhaps we should just always clear the cache, as changes are the DB query will be more expensive than the redis hit anyway.

wagtailio/project_styleguide/templates/patterns/includes/menu_main.html
```diff
@@ -1,42 +1,44 @@
...
-                        {% endwith %}
-                    </li>
-                </ul>
-            </li>
-        {% endfor %}
-    </ul>
-</nav>
+{% load wagtailcore_tags cache %}
+{% wagtail_site as current_site %}
+{% cache 900 primarynav current_site.pk is_pattern_library %}
```
RealOrangeOne, line 3: Question: Should we vary this on `request.is_preview` too? Means preview requests won't get cached alongside normal content, although perhaps not necessary here as its content is unlikely to vary in a preview request.

wagtailio/navigation/models.py
zerolab: I would rather not as we support multiple nav versions, so only the one that needs purging should trigger this

wagtailio/project_styleguide/templates/patterns/includes/menu_main.html
zerolab: Was thinking to defer that to the point where we actually do previews for the nav snippets. But there's no time like now. Great shout

wagtailio/settings/base.py
zerolab: huh, that got lost somehow. re-added. thanks

wagtailio/navigation/models.py
zerolab: poor man's debugger 😂
(context: PyCharm 2022.2 foobared Docker support in some interesting ways)

wagtailio/navigation/models.py
zerolab: perhaps. debated this, and landed on the straighforward implementation here.
happy to be convinced otherwise

wagtailio/navigation/models.py
```diff
@@ -19,6 +21,12 @@ class FooterMenu(models.Model):
...
+            cache.delete(make_template_fragment_key("footernav"))
```
RealOrangeOne, line 28: Issue: This needs a `vary_on` argument

wagtailio/navigation/models.py
```diff
@@ -39,6 +46,12 @@ class MainMenu(ClusterableModel):
         FieldPanel("menu_sections", classname="collapsible"),
     ]
 
+    def save(self, **kwargs):
+        super().save(**kwargs)
+
+        if NavigationSettings.objects.filter(main_navigation=self).exists():
+            cache.delete(make_template_fragment_key("primarynav"))
```
RealOrangeOne, line 53: Issue: This needs a `vary_on` argument

wagtailio/navigation/models.py
```diff
@@ -75,3 +88,12 @@ class NavigationSettings(BaseSiteSetting, ClusterableModel):
         FieldPanel("main_navigation"),
         FieldPanel("footer_navigation"),
     ]
+
+    def save(self, **kwargs):
+        super().save(**kwargs)
+
+        keys = [
+            make_template_fragment_key(key, vary_on=[self.site.pk])
+            for key in ["primarynav", "footernav"]
```
RealOrangeOne, line 97: Issue: This also needs to vary on `request.is_preview`

```suggestion
            make_template_fragment_key(key, vary_on=[self.site.pk, True, False])
            for key in ["primarynav", "footernav"]
```

zerolab: should be `[self.site.pk, False, False]` in that we want to clear the non-preview fragment

RealOrangeOne (approved):

PR #289, opened 2022-10-20
Fix get started block links

zerolab:
This PR fixes the links in the "Get started" block, and updates the nav to make use of `pageurl`


zerolab:
@RealOrangeOne you will like this one

RealOrangeOne (approved):
I do indeed :heart_eyes:

PR #291, opened 2022-10-28
Fix modal positioning and contrast issues

laymonage:
## Description

This fixes the z-index of the modal to avoid overlapping with the main content. Not sure whether the page is supposed to be scrollable when a modal is open, but I still keep the current behaviour in this PR.

I also don't know whether the modal is supposed to be off-center (a bit slightly to the top), but I made it centered in this PR.

### Before

https://user-images.githubusercontent.com/6379424/198681142-31660e9d-bea3-40b6-b672-d5e5ca598ce1.mov

### After

https://user-images.githubusercontent.com/6379424/198682446-ae0749ff-5ba0-4fd5-bef6-5000124fc0d9.mov

### DIFF
--- a/wagtailio/static/sass/components/_modal.scss
+++ b/wagtailio/static/sass/components/_modal.scss
@@ -96,6 +96,12 @@
     &__close-button {
         display: block;
         margin-left: auto;
+
+        svg {
+            @media (forced-colors: active) {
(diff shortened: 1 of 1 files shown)


thibaudcolas (approved):
Looking spot on to me!

PR #294, opened 2022-10-31
Add `<legend>` element to `<fieldset>` for feature index block

laymonage:
Each `<fieldset>` element should have a corresponding `<legend>` element to describe what the set of fields is about.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/fieldset

Not sure if we'd like to use `{% trans %}` or put it [in the block itself](https://github.com/wagtail/wagtail.org/blob/next/wagtailio/features/blocks.py#L60) (as a `heading`  `CharBlock` maybe). However, the block currently seems to be quite specific to "Features" (instead of being a generic categorised block), so I think it's fine to put it in the template.

### DIFF
--- a/wagtailio/project_styleguide/templates/patterns/components/streamfields/features/feature_index_block.html
+++ b/wagtailio/project_styleguide/templates/patterns/components/streamfields/features/feature_index_block.html
@@ -6,7 +6,7 @@
         <legend class="u-sr-only">{% trans 'Filter features by category' %}</legend>
         <div class="features__filter-wrap">
             <input type="checkbox" id="all" name="all" data-feature-filter-all checked>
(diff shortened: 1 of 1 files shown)


thibaudcolas (changes requested):
For screen reader users this is spot on, but for sighted users I think we should either hide the text (`sr-only` class) or add extra spacing between the legend and the checkboxes below.

On smaller viewport widths in particular, the text is partially obscured by the checkboxes below.

laymonage:
Whoops, sorry I missed the `u-sr-only` in the PR somehow! I've added it now, thanks!

thibaudcolas (approved):

PR #296, opened 2022-11-02
Re-add missing This Week in Wagtail logo

laymonage:
All the "This Week in Wagtail" issues throw a 500 error right now, e.g. https://wagtail-org-production.herokuapp.com/this-week-in-wagtail/issue-115/

I believe this is because the static image `this-week-in-wagtail.png` has been removed in https://github.com/wagtail/wagtail.org/commit/8305b801a299bc21fab563ff60786290ce07a1d3, but it's still referenced in the template:

https://github.com/wagtail/wagtail.org/blob/2242e6d9359d560c0958a2fe78add81b32fe6bd1/wagtailio/newsletter/templates/newsletter/newsletter_page.html#L899

If I'm not mistaken, there were some discussions about renaming this to something else. The website design suggests "The Wagtail Journal", but no actual designs for the newsletter itself or a new logo seems to exist. I think until that decision and the new designs have been made, we should at least fix the pages by re-adding the image.


zerolab:
@laymonage did you run the image through ImageOptim or similar optimizer? if not, could you? otherwise LGTM

vossisboss:
@laymonage Just FYI, to the best of my knowledge, we are keeping the name. The Wagtail Journal was a project Dex was managing that dropped off when we all got really busy.

PR #299, opened 2022-11-02
Use the 'is not' operator

allcaps:
It is needlessly complex to invert the result of a boolean comparison. The opposite comparison should be made instead.


zerolab (approved):
:shipit:

PR #301, opened 2022-11-03
Implement focus trap on mobile navigation menu

laymonage:
This PR adds a focus trap to the navigation menu on mobile devices to prevent tabbing outside of the menu and thus focusing on elements that are covered by the menu.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,6 +9,7 @@
       "version": "1.0.0",
       "license": "ISC",
       "dependencies": {
+        "focus-trap": "^7.0.0",
         "micromodal": "^0.4.10"
       },
       "devDependencies": {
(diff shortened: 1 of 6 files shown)


thibaudcolas (changes requested):
I’m facing two issues with this as-is:

- When I attempt to tab past the toggle without opening the menu, either my tabbing gets blocked on the first item within (before your most recent commit), or the tabbing cycles from the toggle to the first item – so I can’t tab past the menu.
- When I open a sub-menu, if the first thing I do is shift-tabbing, this won’t move focus. Shift-tabbing after having tabbed forward first works as expected

I believe I was able to resolve the first issue by only running `this.focusTrap.bindEventListeners();` inside the mobile menu’s `open` function for the top-level `MobileMenu`. If we did do this, we’d need a corresponding function to remove the event listeners on close.

I didn’t investigate the second issue further.

---

All in all, I think we should reconsider having a custom implementation of the focus trap. This is a relatively complex interaction that we might be able to make work for now but I’m worried we won’t necessarily be able to maintain this in the long run.

While we’re at it I think we could also reconsider how sub items work inside the mobile menu. They work as-is but introduce quite a bit of complexity (why does pressing Esc not close the whole menu?). All of this would likely be much simpler if those sub-menus were just `<detail>` elements expanding underneath like an accordion.

laymonage:
@thibaudcolas I've reworked the implementation to use the [focus-trap](https://github.com/focus-trap/focus-trap) library. I agree our menu implementation could be much simpler but I think we could refactor it in a separate PR post-launch.

thibaudcolas (approved):
Solid!

PR #310, opened 2022-11-07
Use a slightly more efficient sitemap

zerolab:
Since we're only outputting the URL and last modified date (`last_published_at` or `latest_revision_created_at`), we don't need the specific objects

(red - before, green - after)
Screenshot 2022-11-07 at 19 40 23


laymonage (approved):
Makes sense to me!

PR #311, opened 2022-11-07
Upgrade to Wagtail 4.1

zerolab:
This PR:

- uses Python 3.10 (could not use 3.11 as poetry is not compatible)
- bumps Wagtail to 4.1
- bumps various packages to latest versions

To note: django-storages 1.13.x drops `AWS_S3_SECURE_URLS`, but have not used it anyway - tested locally with the prod S3 setup and it worked without a hitch


laymonage (approved):
Tested this locally and everything seems to work fine. Thanks!

pyproject.toml
```diff
@@ -24,25 +24,25 @@ django-modelcluster = "~6.0"
...
-wagtailmedia = "~0.11.0"
+wagtailmedia = "~0.12.0"
 whitenoise = "~6.2.0"
 django-redis = "~5.2"
 
 [tool.poetry.group.dev.dependencies]
-black = "22.8.0"
+black = "22.10.0"
 djhtml = "1.5.1"
 fabric3 = "1.14.post1"  # TODO switch to fabric/invoke
```
laymonage, line 47: Can we upgrade this as well? I just noticed `fab` commands are not working because there's an `from collections import Mapping` in Fabric that should be changed to `from collections.abc import Mapping` for Python 3.10

laymonage (changes requested):

pyproject.toml
zerolab: hmm, the fabric upgrade will need more sweeping changes.

In the meantime, we could use https://github.com/mathiasertl/fabric/pull/57

laymonage: Ah, that's unfortunate. Would be nice if we didn't have to use a fork, but if the changes needed to properly upgrade are more involved, I guess it wouldn't hurt to switch to a fork. Not sure how far behind/ahead that branch in the `1.14.post1` version we use, though.

zerolab: it is quite up to date -- https://github.com/mathiasertl/fabric/compare/5913488a95b0318de0a862e7e181217d8a671641..HEAD

laymonage (approved):
`fab` works now, not sure if anything else breaks but hopefully not, I think this looks good!

Issue #308, opened 2022-11-04
Primary nav sub-menus should be dismiss-able with the Esc key


thibaudcolas:
Related: #307. We don’t use a focus trap for those sub-menus, but there should nonetheless be a way to dismiss them with the keyboard.

See https://www.w3.org/WAI/ARIA/apg/example-index/disclosure/disclosure-navigation.html as an example.

thibaudcolas:
Fixed in #314.

Issue #307, opened 2022-11-04
Primary nav toggles should use buttons instead of links, with correct ARIA markup


thibaudcolas:
The main menu’s `data-open-subnav` elements should be `<button>` elements instead of anchors so they are correctly announced to screen reader users.

They seem to follow a [disclosure](https://www.w3.org/WAI/ARIA/apg/patterns/disclosure/) widget pattern, so also shouldn’t use `aria-haspopup`, which is intended for `dialog`, `menu` and other advanced ARIA roles, which our implementation doesn’t use.

Issue #306, opened 2022-11-04
Nav CTA is cropped within mobile menu


thibaudcolas:
Within the mobile menu’s sub-items, all `nav-cta` components at the bottom of each section are only half-visible when scrolling all the way to the end of the menu.

<img width="1425" alt="Screenshot 2023-02-26 at 10 32 15 AM" src="https://user-images.githubusercontent.com/97099574/221393331-a23d1467-9e2f-46dc-83cf-1155ddfe102b.png">

vossisboss:
It looks like removing the Ukraine banner has solved this issue. We were unable to reproduce this item and so we're going to close this issue for now.

Issue #305, opened 2022-11-04
Insufficient color contrast for primary menu chevron icons


thibaudcolas:
Images that aren’t decorative like those icons need a color contrast of at least 3:1. Currently `primary-nav__icon` elements use #c8c8c8 on white.

thibaudcolas:
Fixed in #314.

PR #314, opened 2022-11-09
Close desktop sub-menus when pressing Esc

thibaudcolas:
Fixes:

- #304 & #306 (same fix)
- #305
- #307
- #308
- …and one item from our accessibility audit (alt text for the video).

There isn’t that much in common between those issues, just grouping them so it’s easier to test everything in one go.


laymonage (approved):
Apart from the scrollable mobile menu thing, I've confirmed the fixes and they all look good to me, thanks!

thibaudcolas:
Merged without the problematic change!

PR #315, opened 2022-11-09
Remove footer survey link

zerolab:
Note: did not remove the css or the clipboard icon, but could do that too if needed


thibaudcolas (approved):
👍 I switched to commenting out the code so we don’t forget how the markup is intended to work if we need this again in the future / so we also remove the code if we definitely don’t.

PR #324, opened 2022-11-11
Template tidy ups

zerolab:
- Drops Facebook/Twitter JS (and a share include)
- Uses the page intro, if set, as a fallback for the meta description tag. e.g. for blog posts


laymonage (approved):
Looks good to me. Would be nice to share the code (e.g. a template tag or something) for `<meta name="description"` and other description meta tags (`<meta name="twitter:description"` `<meta property="og:description"`), but we're not going to need it for any other stuff anyway, so this is fine for me

zerolab:
Dropped the twitter.js include too and rebased.

PR #333, opened 2022-11-14
Ensure 404 and 500 views return correct status codes

RealOrangeOne:
When implementing custom 404 / 500 views, they still need to return the correct status codes

This may be causing the issues we're seeing with redirects not working.

### DIFF
--- a/wagtailio/utils/views.py
+++ b/wagtailio/utils/views.py
@@ -18,8 +18,8 @@ def robots(request):
 
 
 def error_404(request, exception=None):
-    return render(request, "patterns/pages/errors/404.html", {})
+    return render(request, "patterns/pages/errors/404.html", {}, status=404)
 
 
 def error_500(request, exception=None):
-    return render(request, "patterns/pages/errors/500.html", {})
+    return render(request, "patterns/pages/errors/500.html", {}, status=500)


laymonage (approved):

wagtailio/utils/views.py
```diff
@@ -18,8 +18,8 @@ def robots(request):
 
 
 def error_404(request, exception=None):
-    return render(request, "patterns/pages/errors/404.html", {})
+    return render(request, "patterns/pages/errors/404.html", {}, status=404)
```
laymonage, line 21:
```suggestion
    status = 404 if exception else 200
    return render(request, "patterns/pages/errors/404.html", {}, status=status)
```

wagtailio/utils/views.py
```diff
@@ -18,8 +18,8 @@ def robots(request):
...
 
 
 def error_500(request, exception=None):
-    return render(request, "patterns/pages/errors/500.html", {})
+    return render(request, "patterns/pages/errors/500.html", {}, status=500)
```
laymonage, line 25:
```suggestion
    status = 500 if exception else 200
    return render(request, "patterns/pages/errors/500.html", {}, status=status)
```

laymonage (commented):
Just realised we should probably set `status` only if the exception exists, so that the page can still be tested using `test404/` and `test500/` urls.

RealOrangeOne:
Good point. Let's see if this resolves the issue, then yes I'll do a subsequent PR

PR #334, opened 2022-11-14
Reuse Django's error views for custom views

RealOrangeOne:
This is an extension and cleanup of #333.

See also https://git.torchbox.com/internal/wagtail-kit/-/blob/master/wagtailkit_repo_name/wagtailkit_repo_name/utils/views.py (Torchbox only)


zerolab (approved):
LGTM

tomusher (approved):

PR #337, opened 2022-11-14
Add logos block to content page

vossisboss:
We need some small adjustments to the Logos block and to the blocks available on the Content page to create a suitable MVP design for the "Who Uses Wagtail" page. 

This PR makes the Logos block available on the Content page type and provides an optional heading that can be used to organise the logos.

Ben E and Ian B from Torchbox want to review the design a bit before we proceed with merging, so I'm submitting this PR as a draft for now.


wagtailio/core/migrations/0056_auto_20221114_2049.py
```diff
@@ -1,30 +1,1589 @@
 # Generated by Django 3.2.16 on 2022-11-14 20:49
```
zerolab, line 1: suggestion: rename this migration. we try to steer away from auto-named migrations

zerolab:
> Ben E and Ian B from Torchbox want to review the design a bit before we proceed with merging, so I'm submitting this PR as a draft for now.

This sounds like a perfect opportunity to get this onto `staging`. Don't have any qualms with the BE changes

vossisboss:
@zerolab I sent Ben and Ian a short walkthrough of the design from my local machine. They approved it, so I think we can proceed to production. But I'm okay with testing on staging first if that's your preference.

wagtailio/static/sass/components/_logos.scss
```diff
@@ -4,11 +4,11 @@
     position: relative;
 
     &__heading {
-        margin-bottom: 20px;
+        margin-bottom: 80px;
```
laymonage, line 7: We might want to reduce this, and maybe increase the margin below the logos instead? Feels a bit weird that the bottom margin of the heading is larger (80px) than the bottom margin between the logos and the next heading (50px), which is a separate group.

image


image

laymonage (commented):
Changes to the blocks look good to me, just a minor issue with the spacing

wagtailio/static/sass/components/_logos.scss
vossisboss: Submitted a commit that sets the margin to `60px` for all views. Please let me know if there's anything else I should fix.

laymonage (approved):

Issue #353, opened 2022-11-23
Use Calibre's image actions to compress images in PRs


zerolab:
https://github.com/calibreapp/image-actions

Previously done manually (e.g. #285)

udbhavsomani:
Hi @zerolab, I am new to contributing to wagtail, shall I proceed with this issue if not resolved already?

zerolab:
Hey @udbhavsomani thank you for expressing your interest. Go for it. The rule is that unless the issue or a comment on the issue explicitly says not to, and there is no-one else who is working on it, then feel free to leave a note that you are working on it, or just open a draft PR. See https://dev.to/lb/ten-tasty-ingredients-for-a-delicious-pull-request-hgc for a lot of useful tips

udbhavsomani:
Hi @zerolab, please review my PR: https://github.com/wagtail/wagtail.org/pull/359

thibaudcolas:
I’d like to propose we _don’t_ do this via GitHub Actions, and instead do it manually. We can document the process to do it manually via something like [Squoosh](https://squoosh.app/) if we want, which should give the same image optimisation benefits without the drawbacks I foresee.

The drawbacks are:

1. As-is, the Calibre action will always attempt to re-optimise images no matter how many times it’s already attempted in the past (see https://github.com/calibreapp/image-actions/issues/92). This is massively wasteful on CPU, which has a clear energy and environmental cost.
2. If the Calibre action decides to optimise an image, then it’s an extra commit with the newly-compressed images. This effectively at least doubles the space images will take in the git history (as many times as there’s been optimisations). Whereas if we optimise more manually we can be more cautious about how many times we do so and whether the savings are worth the bloat in the git history.

The possible benefits also seem way too small to me:

- There are 8 images currently in the repo that could benefit from this. None of them are particularly prone to change
- For any new work on the site, we’d want to support dark mode, which almost entirely rules out usage of raster images for styling purposes (it’s much easier to theme SVGs).

So – if we want to improve on image size, my preference would be to:

1. Convert all raster images we can to SVGs. For example we could push harder on SVG favicons.
2. Document how we optimise the raster images we keep

PR #359, opened 2023-01-02
Added Calibre's image actions to compress images in PRs

udbhavsomani:
Tested by adding 3 different types of images.


.github/workflows/calibreapp-image-actions.yml
```diff
@@ -0,0 +1,22 @@
...
+      - '**.webp'
+jobs:
+  build:
+    # Only run on Pull Requests within the same repository, and not from forks.
+    # if: github.event.pull_request.head.repo.full_name == github.repository
+    name: calibreapp/image-actions
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v2
```
zerolab, line 17:
```suggestion
        uses: actions/checkout@v3
```

https://github.com/actions/checkout

.github/workflows/calibreapp-image-actions.yml
```diff
@@ -0,0 +1,22 @@
...
+  pull_request:
+    paths:
+      - '**.jpg'
+      - '**.jpeg'
+      - '**.png'
+      - '**.webp'
+jobs:
+  build:
+    # Only run on Pull Requests within the same repository, and not from forks.
+    # if: github.event.pull_request.head.repo.full_name == github.repository
```
zerolab, line 12: This is needed because of the limitations described in https://github.com/calibreapp/image-actions#process-pull-requests-from-forked-repositories

so https://github.com/calibreapp/image-actions#combined-workflow is probably the best

zerolab (changes requested):
Thanks for this @udbhavsomani, added a couple of comments

.github/workflows/calibreapp-image-actions.yml
udbhavsomani: @zerolab sure, I'll uncomment this, do you also want me to add cronjobs?

zerolab: https://github.com/calibreapp/image-actions#combined-workflow is the one that we want

thibaudcolas:
@udbhavsomani thank you for giving this a go! I see there hasn’t been much movement on this in the last nine months, so I’ll close this now.

Personally I don’t think those kinds of GitHub Actions are a good idea at all, but that’s better discussed in #353.

PR #361, opened 2023-01-23
Updating email styles and logo for This Week in Wagtail

vossisboss:
**Purpose of this PR:** To update the styles for the This Week in Wagtail newsletter to match the current styling of the webpage redesign.

**NOTE:** The HTML template for the newsletter page involves email HTML, which is why there are a lot of tables in the code. The template Karl chose for the email back when he set this up is a Bootstrap-like responsive email template called Ink, which ultimately became [Foundation for Email](https://github.com/foundation/foundation-emails). The Foundation project has been inactive for a while and there is no documentation available. We should consider switching to a newer responsive email template or framework. But this code can still be used while we look for a better option. I would like to go ahead and focus on making these styles work for now so we can resurrect the newsletter.


laymonage:
I haven't reviewed the PR yet, but if we're looking to change the email CSS framework, [mailwind](https://github.com/soheilpro/mailwind) might be worth looking into.

cnk:
Looks good to me!

PR #368, opened 2023-03-08
Upgrade to Wagtail4.2

### LINKED ISSUE #367: Implementation issues with meta tags for blog posts
See the HTML of a [recent blog post](https://wagtail.org/blog/contributors-webinar-plans-for-stimulus-in-wagtail/):

- `og:url` has one too many `https://`
- `twitter:image` should be a URL, not a path
- `og:image` should be a URL, not a path

```html
<meta property="og:url" content="https://https://wagtail.org/blog/contributors-webinar-plans-for-stimulus-in-wagtail/" />
<meta name="twitter:image" content="/static/img/default-sharing-image.488db1f9c279.png" />
<meta property="og:image" content="/static/img/default-sharing-image.488db1f9c279.png" />
```

Template: https://github.com/wagtail/wagtail.org/blob/1ef5ac08a4d7e8c04e32bb1e68fdf089d84a890c/wagtailio/project_styleguide/templates/patterns/base_page.html#L36

I believe one `https://` needs removing, and the default sharing images need `{{ site.root_url }}` added.

zerolab:
- Wagtail 4.2
- Django 4.1
- Python 3.11
- Poetry 1.4.0
- latest backend linting versions

and fix #367


.pre-commit-config.yaml
```diff
@@ -1,28 +1,28 @@
 default_language_version:
   node: 18.7.0
-  python: python3.10
+  python: python3.11
 repos:
   - repo: https://github.com/psf/black
-    rev: 22.10.0
+    rev: 23.1.0
```
thibaudcolas, line 6: This and all below – looks like we’ll need to do the same upgrades to versions defined in Poetry?

wagtailio/settings/base.py
```diff
@@ -123,7 +123,6 @@
 LANGUAGE_CODE = "en-gb"
 TIME_ZONE = "UTC"
 USE_I18N = True
-USE_L10N = True
```
thibaudcolas, line 126: I think this is still technically needed. Imagine we don’t need this in [Django 4.0 and up](https://docs.djangoproject.com/en/4.1/ref/settings/#use-l10n), but right now the PR doesn’t actually update to Django 4.1?

pyproject.toml
```diff
@@ -12,33 +12,28 @@ requires = ["poetry-core"]
 build-backend = "poetry.core.masonry.api"
 
 [tool.poetry.dependencies]
-python = "^3.10"
-beautifulsoup4 = "~4.9.3"
-boto3 = "~1.26.3"
-dj-database-url = "~1.0.0"
+python = "^3.11"
 django = "~3.2"
```
thibaudcolas, line 16: Did you intend to upgrade Django as well?

wagtailio/images/migrations/0013_wagtail_42.py
```diff
@@ -0,0 +1,33 @@
+# Generated by Django 3.2.18 on 2023-03-09 18:37
```
thibaudcolas, line 1: If I understand https://docs.wagtail.org/en/stable/releases/4.2.html#wagtail-specific-image-field-wagtailimagefield, there won’t be any testing for me to do here as long as the migration runs (it does). So moving on.

poetry.lock
```diff
@@ -1,18 +1,28 @@
...
 name = "anyascii"
 version = "0.3.1"
 description = "Unicode to ASCII transliteration"
 category = "main"
 optional = false
 python-versions = ">=3.3"
+files = [
+    {file = "anyascii-0.3.1-py3-none-any.whl", hash = "sha256:8707d3185017435933360462a65e2c70a4818490745804f38a5ca55e59eb56a0"},
+    {file = "anyascii-0.3.1.tar.gz", hash = "sha256:dedf57728206e286c91eed7c759505a5e45c8cd01367dd40c2f7248bb15c11f6"},
+]
```
thibaudcolas, line 13: Hmm, this metadata will be super hard to review since it looks like Poetry has changed their file format :/ I guess I’ll just trust the process.

pyproject.toml
zerolab: 🤦‍♂️ yup. will take care of it in the afternoon

pyproject.toml
```diff
@@ -37,12 +37,12 @@ whitenoise = "~6.3"
...
+djhtml = "3.0.4"
 fabric3 = { git = "https://github.com/wmorgan6796/fabric.git", rev = "77513f3c" }  # TODO switch to fabric/invoke
-flake8 = "5.0.4"
+flake8 = "6.0.0"
 flake8-assertive = "2.1.0"
 flake8-blind-except = "0.2.1"
-flake8-comprehensions = "3.10.0"
-isort = "5.10.1"
+flake8-comprehensions = "3.10.1"
+isort = "5.12.0"
```
zerolab, line 47: @thibaudcolas I am very much in favour of dropping these and deferring all to pre-commit. What do you think?

pyproject.toml
thibaudcolas: 👍 can always do it in the next upgrade cycle too. Django 3.2 is still LTS.

pyproject.toml
thibaudcolas: I’m happy with the current setup because I can configure my IDE (VS Code) to use the correct versions of those tools via Poetry’s `poetry config virtualenvs.in-project true`. If there is an equivalent way to do this with pre-commit I’m happy too.

PR #369, opened 2023-03-09
Upgrade to wagtail 4.2

victoriachan:
Ticket: https://torchbox.monday.com/boards/1130687432/pulses/1159465613

Note that with this upgrade, Comments within InlinePanel will be removed. ([see](https://docs.wagtail.org/en/latest/releases/4.2.html#comments-within-inlinepanel-not-supported))


zerolab:
@victoriachan thank you for this. I did open #368 yesterday
The PR updates a few other things (like packages) while at it

victoriachan:
Thanks @zerolab . I should have checked first!

PR #375, opened 2023-04-06
Updates to the streamfield spacing mixin and subsequent adjustments

vossisboss:
The spacing between StreamField blocks was too large and making our blog pages look a little funny. I initially removed the media query in `@mixin sf-spacing()` but that broke the spacing around the hero areas and other elements. So I updated `sf-spacing` to include a variable `$multiple` that could be used to adjust spacing for individual elements in multiples of 25px. Then I updated all selectors that use that mixin according to the amount of space needed for each element.

I also removed a media-query the `.icon-bullet` selector since that needed adjustment as well.

Please let me know if there's anything I can do to improve this adjustment to the spacing.

### DIFF
--- a/wagtailio/static/sass/abstracts/_mixins.scss
+++ b/wagtailio/static/sass/abstracts/_mixins.scss
@@ -30,12 +30,8 @@
 }
 
 // Streamfield spacing
-@mixin sf-spacing() {
-    margin-bottom: 50px;
-
-    @include media-query(large) {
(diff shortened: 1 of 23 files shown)


estyxx (approved):
LGTM

estyxx:
If there is any fontend dev (more experienced than me) that can have a look at this change would be nice. Otherwise seems ok to me...

Issue #346, opened 2022-11-18
Block cleanup on BlogPage


vossisboss:
There are two main things that need to happen for this issue to be addressed. First, old block types that we are no longer using anymore need to be removed from the StreamField for `BlogPage` so that people won't use them anymore. Second, certain blocks from `ContentPage` need to be added to `BlogPage` so people can use them in their blogs as well.

The following blocks need to be removed from `BlogPage` so that people don't use them anymore:

- Intro
- Textimage
- Colourtext
- Calltoaction
- Stats
- Tripleimage
- Backers
- Image caption

The following blocks from `ContentPage` need to be available in `BlogPage`:

- Get Started block
- Sign up form
- Highlight
- Teaser

When these blocks are added to `BlogPage`, we'll need to confirm that they display appropriately with the frontend styles used for `BlogPage`.

fyunusa:
Hi, @vossisboss i'll be taking up on these. Hoping to send a good report soon

vossisboss:
Thank you for giving this a go @fyunusa. Do let us know if you have any questions.

fyunusa:
Hi @vossisboss , i have been able to remove the blocks from **blogPage** and added the blocks needed on **contentPage**.
<img width="1280" alt="Screenshot 2022-11-20 at 12 44 47 PM" src="https://user-images.githubusercontent.com/55334829/202900234-b4398560-fc4f-4f52-b88e-4fd789481685.png">

### But then i have few things to clarify on.

**First** 
The the blogPage storyBlock inherits from the utils storyBlock and extends it with a method for the **standalone_cta block**, which makes up the current **blogPage storyBlock** , using same approach i was able to extend the blogPage storyBlock with new Blocks needed:
- Get Started block
- Sign up form
- Highlight
- Teaser
<img width="713" alt="Screenshot 2022-11-20 at 12 45 20 PM" src="https://user-images.githubusercontent.com/55334829/202900553-84725fc4-9e3b-47fe-ae71-af08458c9981.png">

**Secondly**
Due to the blogPage storyBlock inherits from the utils storyBlock with my little knowledge of the system 
**The current codebase shows that these blocks are to be removed from the utils storyBlocks**
`imagecaption = ImageAndCaptionBlock(label="Image caption")  # to be removed
    textimage = TextAndImageBlock(icon="image")  # uses text_and_media_block.html
    colourtext = BackgroundColourTextBlock(icon="pilcrow")  # to be removed
[...]

PR #382, opened 2023-05-18
Cleaning up the blog blocks

vossisboss:
This PR is for cleaning up the blog blocks. We have a lot of blocks that were slated to be removed that are cluttering up the block chooser for StreamField. We also had blocks on the standard Content page that needed to be accessible in the blog.

I added three blocks to  `StoryBlock`: Teaser Block, Get Started, and Sign-up form

I created two new groups called "DO NOT USE" and "CTA options" to organise the blocks. I put all the old unsupported blocks in the "DO NOT USE" groups for now since I'm unsure how removing the blocks from `StoryBlock` will affect existing block content in the database and I suspect cleaning those blocks up properly will also involve making sure the existing content has a fallback-display option.

If there are any good examples of retiring blocks, let me know. Otherwise, I think this solution will take care of most of our immediate needs for the blog.


wagtailio/utils/blocks.py
```diff
@@ -222,12 +226,12 @@ class StoryBlock(StreamBlock):
...
-    colourtext = BackgroundColourTextBlock(icon="pilcrow")  # to be removed
-    calltoaction = CallToActionBlock(icon="pilcrow")  # to be removed
-    tripleimage = TripleImageBlock(icon="image")  # to be removed
-    stats = ListBlock(StatBlock(icon="code"))  # to be removed
+    imagecaption = ImageAndCaptionBlock(label="Image caption", group= "DO NOT USE")  # to be removed
+    textimage = TextAndImageBlock(icon="image", group= "DO NOT USE")  # uses text_and_media_block.html
+    colourtext = BackgroundColourTextBlock(icon="pilcrow", group= "DO NOT USE")  # to be removed
+    calltoaction = CallToActionBlock(icon="pilcrow", group= "DO NOT USE")  # to be removed
+    tripleimage = TripleImageBlock(icon="image", group= "DO NOT USE")  # to be removed
+    stats = ListBlock(StatBlock(icon="code"), group= "DO NOT USE")  # to be removed
```
zerolab, line 234: I am pretty sure I cleaned up the pages that were using these blocks, hence the `# to be removed` comments.

Will run a check tomorrow will update this

zerolab:
@vossisboss #346 includes the Highlight block in the "to add" list. is that no longer needed?

wagtailio/utils/blocks.py
zerolab: I can confirm no pages use these anymore, so please remove with prejudice.
Don't forget to remove the actual block definitions and the corresponding templates.

Locally, I suggest running:
```
./manage.py migrate blog 0011
./manage.py migrate standardpage 0006
```
removing the migrations in this PR, then regenerating the migration once the cleanup is done

zerolab (changes requested):
The "DO NOT USE" blocks can be removed, so can you please do so?

Also, CI is failing due to linting issue. We use pre-commit, so `pre-commit install` in the project root, then `git ls-files --others --cached --exclude-standard | xargs pre-commit run black --files` and `git ls-files --others --cached --exclude-standard | xargs pre-commit run isort --files` in the first instance, and then it should just work

vossisboss:
@zerolab I think have cleaned up all associated templates and block definitions. 

I tried removing the block definition for `BackgroundColourChoiceBlock` as well since I don't believe we need that one anymore either, but the test server and the automatic migrations wouldn't run if I removed that one. There was an error `AttributeError: module 'wagtailio.utils.blocks' has no attribute 'BackgroundColourChoiceBlock' `

I'm not 100% sure why removing that one is different compared to the others that were removed. How did you check to see if the other blocks were used before you removed them? Checking the database?

zerolab:
BackgroundColourChoiceBlock is references in previous migrations. Let's leave it for now
Go for staging

vossisboss:
So @zerolab, the staging branch hadn't been updated for a while. So I created a new one from `main` and merged my changes to the new staging branch and deployed that branch to staging. I clearly borked something though because I'm getting a lovely 400 error. Any recommendations on which logs I should poke through for troubleshooting?

zerolab:
I think at this point we need to bring staging on par with prod. Will see what I can do about it

vossisboss:
@zerolab Is there a reason we don't duplicate the data from production in staging beyond not wanting to pay to have the same data in two places? I see we have a smaller database for the staging app, so clearly we don't want to go nuts. But I think it would make testing things in staging a bit smoother if we at least had the homepage, the main page structure, some key snippets, and a couple of the blogs.

Would anything blow up terribly if I used one of the `make` commands to pull the production data and media into staging? I could prune the amount of content down after the initial pull so that we don't have _everything_ in there.

zerolab:
It should be ok to do so with the caveat we anonymise and non-essential users

vossisboss:
All right. I did some test content and the blocks appear to be working they way they were intended to on staging. Unless there are any other tests or checks I should do, I would like to push this to production.

zerolab (approved):

PR #389, opened 2023-07-11
Updated Dockerfile to use AWS CLI v2

vossisboss:
The `make build` command was failing because it seems the AWS CLI file we were using is no longer available for download. I updated the commands in the Dockerfile to use AWS CLI version 2.

I tested the updated code with a fresh build in Docker on my local machine. I would appreciate someone else who has production access testing that this update works on their machine as well before merging this commit.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -77,10 +77,10 @@ RUN apt-get install -y nodejs
 RUN curl https://cli-assets.heroku.com/install.sh | sh
 
 # Install AWS CLI
-RUN curl "https://s3.amazonaws.com/aws-cli/awscli-bundle.zip" -o "/tmp/awscli-bundle.zip" && \
-    unzip /tmp/awscli-bundle.zip -d /tmp && \
-    /tmp/awscli-bundle/install -i /usr/local/aws -b /usr/local/bin/aws && \
-    rm /tmp/awscli-bundle.zip && rm -r /tmp/awscli-bundle
+RUN curl "https://awscli.amazonaws.com/awscli-exe-linux-aarch64.zip" -o "awscliv2.zip" && \
+    unzip awscliv2.zip && \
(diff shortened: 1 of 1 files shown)


cnk:
I tried this out in my Intel-based Mac. The version currently on main (using https://s3.amazonaws.com/aws-cli/awscli-bundle.zip) does not build at all. The new version (using https://awscli.amazonaws.com/awscli-exe-linux-aarch64.zip) builds for me but I don't appear to be able to use the `aws` CLI.

```
wagtailio@f3a3a21c0705:/app$ aws --help
qemu-aarch64: Could not open '/lib/ld-linux-aarch64.so.1': No such file or directory
wagtailio@f3a3a21c0705:/app$ which aws
/usr/local/bin/aws
```

To decide if this PR is the correct solution I think we need to know 
1. What architecture is correct for the production deploy to Heroku (and is creating the correct production container taken care of by a GitHub actions pipeline)?
2. What architecture(s) do we want to support in development. Do we need to support both M1/ARM and Intel containers for development? If not, then we need to be clearer about which architectures can use the Docker Compose installation - and steer everyone else to ? Vagrant ? or local installation?

RealOrangeOne:
The best way to handle this is to use atrictly amd64 (x86) containers, and instruct macOS to run the containers in a compatibility layer (which is very simple). The issue here is that the "aarch64" builds are ARM only, so won't work anywhere else.

nicklee:
This has fixed my build on apple silicion 👍

vossisboss:
@RealOrangeOne You said in a separate discussion that AWS-CLI is not strictly necessary for our containers. So should we just remove AWS-CLI?

thibaudcolas:
Closing in favour of #392 as Sage has approved that

PR #402, opened 2023-08-25
Minor fixes

### LINKED ISSUE #399: Incorrect `text-shadow` applied to newsletter signup sub-heading in footer
The `text-shadow` added in #397 seems to inadvertently cause an issue with the corresponding element in the footer:

image

The issue also exists in dark mode, but it is less prominent as the footer background colour is closer to `var(--color--background)`:

image

### LINKED ISSUE #400: Author profile picture in related blog posts is not shown correctly in dark mode
If a blog post has a related post, the related post author's profile picture is not shown correctly in dark mode:

image

This isn't an issue in light mode:

image

laymonage:
- Fixes #399 
- Fixes #400

image

image

image

image

## Regression check

From what I tested, this doesn't seem to interfere with the CSS that was added that caused this bug.

The `text-shadow` added in #397 still works for its intended use case:
image

The added padding for the `logo-card` image still works too:
image

### DIFF
--- a/wagtailio/static/sass/components/_logo-card.scss
+++ b/wagtailio/static/sass/components/_logo-card.scss
@@ -71,5 +71,9 @@
             padding: 10px 20px;
             background-color: $color--white;
         }
+
+        #{$root}__author-wrap #{$root}__image {
+            padding: 0;
+        }
     }
 }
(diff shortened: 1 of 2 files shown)


laymonage:
Merging due to urgency ✌️

PR #414, opened 2023-10-06
Don't index search pages

RealOrangeOne:


wagtailio/project_styleguide/templates/patterns/pages/search_results/search_results.html
```diff
@@ -5,6 +5,10 @@
 {% else %}No results found{% endif %}
 {% endblock title %}
 
+{% block extra_css %}
```
zerolab, line 8: why not 

````
{% block meta_tags %}
    {{ block.super }}

    <meta name="robots" content="noindex">
{% endblock %}

thibaudcolas (approved):
@RealOrangeOne please include a description in the future so we have a track record of the rationale?

I’m not clear why we wouldn’t want any search result page to be indexed, but I assume there’s a reason so will merge this now with Dan’s suggested changes.

PR #416, opened 2023-10-08
changed font-size from px to rem/#322

oly-baby:


thibaudcolas:
@oly-baby 👋 I’m not entirely sure how you’re going about this, it looks like you’re making changes to the project’s files directly in GitHub without checking what the styles look like?

This isn’t going to work, 1px and 1rem aren’t the same thing, and I think there’s a case to refactor the code while we’re here rather than just change the units. 

If you’ve never worked with `rem` before I’d suggest you spend some time getting up to speed before spending more time on this.

oly-baby:
Alright mentor

I understand now, I have to view the code output to change it.

But why is the project closed?

Issue #409, opened 2023-10-02
Text disappears when hovering on the main teaser in dropdown menus


benenright:
https://github.com/wagtail/wagtail.org/assets/4907819/17ba2792-d537-45bc-92e7-8a6e76d1f0e3

Not an issue in Dark mode.

nicklee:
This has been fixed via https://github.com/wagtail/wagtail.org/pull/425 (waiting for next deployment)

PR #420, opened 2023-10-09
Added styling on nav-cta__subheading to prevent it's disappearance on hovering

maureenblack:
#409


thibaudcolas:
@maureenblack thank you for giving this a go, unfortunately this has already been fixed in #425.

maureenblack:
Yeah I did it a long time ago.

Now what can I do instead?
I have a feeling of I asked first before working on this one we'd have both had a happy ending..😄

PR #421, opened 2023-10-09
Update README.md

khushishikhu:
It creates a confusion for a new contributor, as what is the purpose of making "superuser" and how to move ahead.

Added an essential step to ease new contributor

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,7 @@ To create a new superuser locally, run:
 ```
 make superuser
 ```
+To login with Wagtail, please redirect to http://localhost:8000/admin/
 
 ### Pulling production data / media
 
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ Then, to start the development server, open a new terminal window and run:
 make runserver
 ```
 
-This will launch `django-admin runserver` in the web container, which serves Wagtail on http://localhost:8000/
+This will launch `django-admin runserver` in the web container, which serves the site on http://localhost:8000/, with the admin at http://localhost:8000/admin/.
 
 To run migrations within Docker you'll need to run:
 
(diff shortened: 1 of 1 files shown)


MugoBrian:
@khushishikhu , this is great but I think the README.md should be kept as simple as possible and clean, since the [docs](https://docs.wagtail.org/en/stable/) extensively explain the getting started process.

I'd like to suggest including the following instead:
"These superuser credentials are essential for accessing the site as an administrator via the /admin route. "

It's worth noting that contributors setting up their codebase in Gitpod are redirected to a different URL, rather than localhost:8000."

MugoBrian:
@khushishikhu, this is great but I think the README.md should be kept as simple as possible and clean, since the [docs](https://docs.wagtail.org/en/stable/) extensively explain the getting started process.

I'd like to suggest including the following instead:
"These superuser credentials are essential for accessing the site as an administrator via the /admin route. "

It's worth noting that contributors setting up their codebase in Gitpod are redirected to a different URL, rather than localhost:8000."

MugoBrian:
Hello @khushishikhu, this is great work, however I believe the README.md should be kept as simple as possible and clean, since the [docs](https://docs.wagtail.org/en/stable/) extensively explain the getting started process.

I'd like to suggest including the following instead:
"These superuser credentials are essential for accessing the site as an administrator via the /admin route. "

It's worth noting that contributors setting up their codebase in Gitpod are redirected to a different URL, rather than localhost:8000."

thibaudcolas (approved):
Thanks @khushishikhu! As @MugoBrian it’s crucial for us those instructions are quite succinct – so I kept what you added but moved it to the line we already have about which page to open to view the site (and now the admin).

PR #423, opened 2023-10-15
Changed the former Twitter logo to the new X logo

Amarajah:
I also added a JPEG of it to the wagtailio/static/img folder


thibaudcolas:
Thank you @Amarajah! Merged in 2ed61638. I don’t think we need to update the static files here, and if we did in the future we’d definitely want to keep the same file name as before.

PR #427, opened 2023-10-20
Improved accessibility of signup form

### LINKED ISSUE #426: Improve accessibility of signup form
Modifications should be made to the newsletter signup form to improve its accessibility and ease of use. For example, 
- When screen reader users navigate a page by h2 and h3, they land on the same signup form. This can be confusing since the signup form is the only content under both headings
- The form is lacking an autocomplete attribute even though it collects personal information about the user
- Semantic HTML element - the fieldset - has been used for styling and layout purpose where a generic element is recommended

praiztech:
This pull request has been created to fix #426 issue.

I've made the following modifications to the signup form:
- Added an autocomplete of email to improve the ease of filling out the form.
- Added distinct focus styling so that keyboard-only users can easily determine when they have navigated to the form.
- Removed the aria-label of Enter your email address and Click to sign up, on email input and button span respectively. The email field `<label>` and sign up button text already provide an accessible name for the respective fields.
- Used the more generic `<div>` for layout of the email field.
- Updated the form's subheading to use a `<p>` instead of an `<h3>`, to improve the experience for screen reader users when they navigate a webpage by headings.
- Updated the subheading of the signup form in the footer to include information about the required input. This would aid users with cognitive disabilities to fill out the form since the email field's visible label disappears when the user types into the field.
- Updated the styling for the email field of the signup form in the footer. Added styles to ensure that the email field takes up all of the available width at a breakpoint of 599px. This allows users to easily see the full email address typed.

### DIFF
--- a/wagtailio/static/sass/abstracts/_mixins.scss
+++ b/wagtailio/static/sass/abstracts/_mixins.scss
@@ -94,6 +94,14 @@
     }
 }
 
+// Places the focus indicator within the element
+@mixin focus-internal() {
+    &:focus-visible {
+        outline-offset: -4px;
+        outline: 2px solid currentColor;
+    }
+}
+
 // z-index
 // Map z-index keys
 @function z-index($key) {
--- a/wagtailio/static/sass/abstracts/_mixins.scss
+++ b/wagtailio/static/sass/abstracts/_mixins.scss
@@ -98,7 +98,6 @@
 @mixin focus-internal() {
     &:focus-visible {
         outline-offset: -4px;
-        outline: 2px solid currentColor;
     }
 }
 
--- a/wagtailio/static/sass/base/_base.scss
+++ b/wagtailio/static/sass/base/_base.scss
@@ -40,7 +40,7 @@ textarea {
     font-family: inherit;
 }
 
-div, fieldset {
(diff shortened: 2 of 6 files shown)


wagtailio/static/sass/abstracts/_mixins.scss
```diff
@@ -94,6 +94,14 @@
     }
 }
 
+// Places the focus indicator within the element
+@mixin focus-internal() {
+    &:focus-visible {
+        outline-offset: -4px;
+        outline: 2px solid currentColor;
```
thibaudcolas, line 101:
```suggestion
```

It seems to me that we could move the focus indicator within, without introducing new styles for it.

wagtailio/static/sass/base/_base.scss
```diff
@@ -40,7 +40,7 @@ textarea {
     font-family: inherit;
 }
 
-fieldset {
+div, fieldset {
```
thibaudcolas, line 43: This is puzzling to me – `div` elements have no border or padding set by default.

```suggestion
fieldset {
```

thibaudcolas (approved):
Thank you @praiztech!

PR #435, opened 2023-10-27
Fixing comparison table accessibility issues (#435). Fix #318

### LINKED ISSUE #318: Comparison table accessibility issues
- If the table has an empty header cell, it shouldn’t be marked up as an empty `<th>` element. Use an empty `<td>` instead.
- Table column headers (first row in bold) should be semantically marked up as `<th scope="col">`
- Table row headers (first cell in bold) should be semantically marked up as `<th scope="row">`

shyusu4:
Fixes #318


thibaudcolas (approved):
Looking and working great, thank you @shyusu4!

Issue #440, opened 2023-10-30
SVG icons are not same size on smaller screen and dark Mode toggle button Accessibility


ShubhamOulkar:
**Description:**
1. SVG icons are not same size for smaller screen. Following image shows subsection from home page.
2. A dark theme toggle in the footer, we must take extra care to make it visible, clearly labeled, and accessible to all users. Position is an important accessibility feature in a location that is more easily discoverable and predictable for users.

**How User will experience ? :** wagtail.org home page is long, you have to scroll all the way down to find dark theme toggle button. Especially on mobile devices it is long scroll. This button is not showing any labels or alternate navigation to dark theme toggle. 
Not having same size icons on smaller screen create contrast in the page contain.  User might scroll down without reading content on the screen.

**Images:**
![Screenshot_20231030-080956_Brave](https://github.com/wagtail/wagtail.org/assets/91728992/1b82c657-f9bd-4187-b514-412d85fe2507)
![Screenshot_20231030-080939_Brave](https://github.com/wagtail/wagtail.org/assets/91728992/49081776-0542-4389-b9f8-07a08853f324)

![footer](https://github.com/wagtail/wagtail.org/assets/91728992/2e382bd1-bc5a-4236-abdf-2f309a5a633d)

![footer_phone](https://github.com/wagtail/wagtail.org/assets/91728992/86c92cf0-8b66-4aef-94a8-2fc4da0e0fd8)

**Why this is accessibility issue ?**
1. _**Visibility and Discoverability issue:**_ Footers are typically located at the bottom of a page and may not be immediately visible or discoverable to all users. Users with mobility or vision impairments may have difficulty finding or accessing the button.

2. **_User Expectations issue:_**  Users often expect theme toggles and important interface options to be located in more prominent and predictable positions, such as the header, settings menu, or navigation bar. Placing it in the footer may confuse users, leading to a poor user experience.

4. _**Context Switching issue on all devices:**_ Users who rely on screen readers may find it disorienting if the dark theme toggle button is in an unexpected location. Proper context and labeling are crucial for users who navigate content through screen readers.

5. _**Contrast and Readability issue on mobile devices:**_ Dark theme toggle buttons often involve a visual change in the user interface. Placing it in the footer may lead to contrast and readability issues, as the footer is usually less prominent and may not provide enough visual separation for the button.

**Here are some common and effective locations to consider:**
[...]

PR #441, opened 2023-11-04
sass code fixes inside footer section: hover over SVG icons effect (Follow us icons)

ShubhamOulkar:
We can set color for SVG elements in two ways: using fill and stroke properties of the style attribute. Know more on [MDN docs](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/fill).

# hover over SVG icons effect on follow us links

Untitled video - Made with Clipchamp

This is a PR for issue [#440](https://github.com/wagtail/wagtail.org/issues/440)

@thibaudcolas  , I am outreachy applicant. could you kindly review it. I want to move on the other fixes.

### DIFF
--- a/wagtailio/static/sass/layout/_footer.scss
+++ b/wagtailio/static/sass/layout/_footer.scss
@@ -172,8 +172,8 @@
         transition: color 0.3s;
 
         &:hover {
-            color: $color--teal-100;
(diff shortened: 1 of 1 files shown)


thibaudcolas (approved):
Thanks @ShubhamOulkar!

PR #442, opened 2023-11-06
Adding LoopingVideoBlock to ContentPage

vossisboss:
Currently, we do have the ability to add videos to the ContentPage type and we need to be able to add the occasional demo video. In this PR, I have:

- Added `LoopingVideoBlock()` to `ContentStoryBlock()`
- Removed autoplay from `LoopingVideoBlock` to reduce the emissions impact of autoplay videos and discourage their use
- Updated the associated migrations

Please let me know if there are any changes that could make this a bit more DRY.


wagtailio/core/blocks.py
```diff
@@ -424,11 +424,6 @@ class Meta:
 class LoopingVideoBlock(blocks.StructBlock):
```
zerolab, line 424: suggestion: rename this to `VideoBlock` to better reflect its nature. 
This would mean the homepage video will need re-adding, unless you do https://docs.wagtail.org/en/stable/advanced_topics/streamfield_migrations.html#streamfield-data-migrations

Also https://github.com/wagtail/wagtail.org/blob/b6f8ea896998ba6fb268ac7b3cb7fb8105da8e67/wagtailio/project_styleguide/templates/patterns/components/streamfields/looping_video_block/looping_video_block.html#L9 needs a tidy up.

zerolab (changes requested):
Left a suggestion for tidy up so we don't accrue tech debt

wagtailio/core/blocks.py
vossisboss: Good idea. I think re-adding the homepage video would be the preferred approach for managing this. If it's just one video, we can definitely do that. Will test it locally first of course.

vossisboss:
@zerolab Ready for another check. When I ran the `make migrations` command, there were no automatically generated migration files despite switching the name of the block. Is this typical? Makes me wonder if I'm doing something wrong.

zerolab (approved):
:shipit:

wagtailio/core/blocks.py
```diff
@@ -421,14 +421,9 @@ class Meta:
         template = "patterns/components/get-started/get-started.html"
 
 
-class LoopingVideoBlock(blocks.StructBlock):
+class VideoBlock(blocks.StructBlock):
     # setting autoplay to True adds 'autoplay', 'loop' & 'muted' attrs to video element
```
helenb, line 425: Suggestion: remove this comment as it is no longer relevant

PR #444, opened 2023-12-08
Install and enable image optimizer binaries

zerolab:
This will help with rendition file sizes all across.

Example using the staging site homepage:
before
after

Dev note: after deployment, run `django-admin wagtail_update_image_renditions` on prod


vossisboss:
I see no reason not to merge this in right away. If nothing else broke on staging, I say let's merge this @zerolab.

PR #445, opened 2023-12-15
First draft for importing Mailchimp campaigns

saevarom:
I've created a first attempt at importing the Mailchimp newsletters. It should only import campaigns that have been sent (have a 'send_time' attribute). And it will only import campaigns that have not been imported before.

You will need to put the Mailchimp API key in the environment as `MAILCHIMP_API_KEY`


saevarom:
I have no idea what is going on with the linting, I get a pass from the pre-commit hook locally, but the github action claims that the template fails the check.

saevarom:
Did you have a chance to take a look at this @vossisboss ?

vossisboss:
@saevarom I did not! I was planning to follow up with you via Slack and I missed this PR somehow. I'll have a look right now. Thank you so much for doing this and for poking me.

vossisboss:
Hello @saevarom. This took longer than I expected because of some trouble getting Docker and Poetry to play well together. Anyway, I tested your code in a local installation and the sync worked pretty well. I wasn't sure if the HTML would look good for the individual pages but it worked really well!

The sync itself worked a little too well though. Seems like almost every campaign with an existing archive item was duplicated (see the screenshots) because none of them have the new field that you added . Perhaps we need to add some date parameters.

The last newsletter in the current archive was sent on 14 July 2023, so for the initial sync, we could filter the campaigns down to any that were sent after 15 July 2023. Then maybe could update the sync to check the past month and match against `mailchimp_campaign_id` to prevent duplicates.

What do you think of that approach? Does that seem reasonable or do you have another idea?

I'm going to email you a test key so that you can run some tests on your own as well as well refine this. It should be in your email inbox shortly after I post this. If you need access to the data and media too, I can poke someone to set that up.

Screenshot 2024-01-13 at 2 11 34 PM
Screenshot 2024-01-13 at 2 11 52 PM

vossisboss:
@saevarom, I'm closing this PR since the new Wagtail newsletter package will likely change how we implement and re-vamp the newsletter page.

PR #446, opened 2023-12-20
Add Showcase page

William-Blackie:
Hey all this PR creates the latest page type; a  Showcase page.

Figma: 
- Light theme: [here](https://www.figma.com/file/uA7pjLNeuSHFr490tfc5HH/Wagtail.org?type=design&node-id=4581-23849&mode=design&t=YM3VwtgB6ze0MFx6-0)
- Dark theme: [here](https://www.figma.com/proto/uA7pjLNeuSHFr490tfc5HH/Wagtail.org?type=design&node-id=4581-22218&viewport=-4346%2C218%2C0.16&t=ch4ZPtlnkrA15MB1-0&scaling=scale-down-width&starting-point-node-id=4581%3A22218&show-proto-sidebar=1&hide-ui=1)
- Prototype: [here](https://www.figma.com/proto/uA7pjLNeuSHFr490tfc5HH/Wagtail.org?type=design&node-id=4581-23849&viewport=-4346%2C218%2C0.16&t=ch4ZPtlnkrA15MB1-0&scaling=scale-down-width&starting-point-node-id=4581%3A23849&show-proto-sidebar=1&hide-ui=1)

What I've done: 
- Created a page listing of related models, with Sector snippets used to filter said listing.
- Create Sector snippets.
- Added HTMX for AJAX functionality of the listing itself as well as using it to create on-click listeners for the cards; this method is more accessible than wrapping each card in an anchor tag.
- Added styling and fixed some existing.
- Added a new SVG to cover this functionality.

Screenshots: 
Mobile: 

Tablet: 

Desktop:


wagtailio/core/models/pages.py
```diff
@@ -130,3 +138,165 @@ class ContentPage(Page, HeroMixin, SocialMediaMixin, CrossPageMixin):
...
+
+    sector = models.ForeignKey(
+        "Sector",
+        models.CASCADE,
+        null=False,
+        blank=False,
+        related_name="+",
+    )
+
+    page = ParentalKey(
```
zerolab, line 258: nitpick: the convention is to have this first. it reduces the cognitive load / scanning time in understanding the setup

wagtailio/core/models/pages.py
```diff
@@ -130,3 +138,165 @@ class ContentPage(Page, HeroMixin, SocialMediaMixin, CrossPageMixin):
...
+    )
+
+    panels = [
+        FieldPanel("title"),
+        FieldPanel("sector"),
+        FieldPanel("description"),
+        FieldPanel("image"),
+        FieldPanel("alt_text"),
+        FieldPanel("logo"),
+        FieldPanel("external_link"),
```
zerolab, line 271: suggestion: wrap the link panels in a `MultiFieldPanel` with some help text around what is required (i.e. you need to enter either an external or internal link). It is the little tweaks that lead to editorial delights.

also, it may be nice if you specified icons for the panels

wagtailio/core/models/pages.py
```diff
@@ -130,3 +138,165 @@ class ContentPage(Page, HeroMixin, SocialMediaMixin, CrossPageMixin):
...
+            raise ValidationError("You must specify an image when using a logo")
+
+    def __str__(self):
+        return self.title
+
+    @property
+    def url(self):
+        if self.external_link:
+            return self.external_link
+        return self.internal_link.url
```
zerolab, line 296: question: is there any chance this could be shifted to the template so as to make use of the request caching (ref: https://docs.wagtail.org/en/stable/advanced_topics/performance.html#page-urls)

wagtailio/project_styleguide/templates/patterns/pages/showcase_page/showcase_page.html
```diff
@@ -0,0 +1,46 @@
...
+                        <option {{ selected }} value="{{ name }}">{{ name }}</option>
+                    {% endfor %}
+                </select>
+            </form>
+        </div>
+
+        {% include "patterns/pages/showcase_page/showcase_page_fragment.html" %}
+
+        <div class="showcase-page__detail">
+            <p>List curated by the Wagtail core team.</p>
```
zerolab, line 40: question: should this be editorial?

wagtailio/core/models/pages.py
```diff
@@ -130,3 +138,165 @@ class ContentPage(Page, HeroMixin, SocialMediaMixin, CrossPageMixin):
...
+        context = super().get_context(request)
+        sectors = {}
+
+        for sector in self._filter_used_sectors():
+            if sector == request.GET.get("sector"):
+                sectors[sector] = "selected"
+            else:
+                sectors[sector] = ""
+
+        context |= {
```
zerolab, line 214: `context.update()` please

wagtailio/core/migrations/0061_create_initial_sectors.py
```diff
@@ -0,0 +1,30 @@
...
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("core", "0060_sector_showcaseitem_showcasepage"),
+    ]
+
+    operations = [
+        migrations.RunPython(
+            code=_populate_sector_data,
+            reverse_code=migrations.RunPython.noop,
```
zerolab, line 28: question: why `noop`?

wagtailio/core/models/pages.py
```diff
@@ -130,3 +138,165 @@ class ContentPage(Page, HeroMixin, SocialMediaMixin, CrossPageMixin):
...
     ]
+
+
+class ShowcasePage(Page, SocialMediaMixin, CrossPageMixin):
+    template = "patterns/pages/showcase_page/showcase_page.html"
+    htmx_template = "patterns/pages/showcase_page/showcase_page_fragment.html"
+
+    parent_page_types = ["core.HomePage"]
+
+    introduction = models.TextField(
```
zerolab, line 149: q: should this be a limited `RichTextField`? if not, wouldn't a `models.CharField` suffice?

wagtailio/core/models/pages.py
```diff
@@ -130,3 +138,165 @@ class ContentPage(Page, HeroMixin, SocialMediaMixin, CrossPageMixin):
...
+            return self.htmx_template
+        else:
+            return self.template
+
+    def serve(self, request):
+        response = super().serve(request)
+        if self._is_htmx_request(request):
+            # We only return the fragment, rather than the whole page.
+            new_url = self.url + "?" + request.GET.urlencode()
+            response.headers["HX-Push"] = new_url
```
zerolab, line 201: You can also add https://htmx.org/headers/hx-push-url/ and avoid the inline script in the template fragment

wagtailio/project_styleguide/templates/patterns/pages/showcase_page/showcase_page_fragment.html
```diff
@@ -0,0 +1,57 @@
...
+    {% for item in showcase_items %}
+        <div hx-on:click="window.location.replace('{{ item.url }}')" class="showcase-card">
+            {% if pattern_library %}
+                {# I've tried everything to get this to render using the pattern-library, #}
+                {# This is a placeholder until a fix is found. #}
+                <img src="https://placehold.co/600x400" width="600" height="400" alt="Some alt" class="showcase-card__image">
+                <img src="https://placehold.co/120x100" width="120" height="100" alt="Some alt" class="showcase-card__logo">
+            {% else %}
+                <picture>
+                    {% image item.image max-600x400 format-webp as webp_image %}
```
zerolab, line 12: If you or someone else reviews #443, we could make use of the new [`picture` template tag](https://docs.wagtail.org/en/stable/releases/5.2.html#responsive-multi-format-images-with-the-picture-tag) in 5.2

wagtailio/project_styleguide/templates/patterns/pages/showcase_page/showcase_page_fragment.html
```diff
@@ -0,0 +1,57 @@
...
+                            height="{{ base_img.height }}"
+                            width="{{ base_img.width }}"
+                            class="showcase-card__logo"
+                            alt="{% firstof item.alt_text item.image.alt %}"/>
+                    </picture>
+                {% endif %}
+            {% endif %}
+            <div class="showcase-card__content">
+                <h2 class="showcase-card__heading">{{ item.title }}</h2>
+                <p class="showcase-card__description">{{ item.description }}</p>
```
zerolab, line 50: Currently the showcase item description can be empty. Is that intended? If so, do we need the empty `p.showcase-card__description` ?

wagtailio/project_styleguide/templates/patterns/pages/showcase_page/showcase_page_fragment.html
```diff
@@ -0,0 +1,57 @@
...
+                <a class="showcase-card__link" href="{{ item.url }}">{{ item.link_text }} <svg class="showcase-card__svg" width="15" height="16"aria-hidden="true">
```
zerolab, line 51:
```suggestion
                <a class="showcase-card__link" href="{{ item.url }}">{{ item.link_text }} <svg class="showcase-card__svg" width="15" height="16" aria-hidden="true">
```

zerolab (changes requested):
First pass review. Left a number of suggestions and questions

wagtailio/core/models/pages.py
```diff
@@ -130,3 +138,165 @@ class ContentPage(Page, HeroMixin, SocialMediaMixin, CrossPageMixin):
         index.SearchField("intro"),
         index.SearchField("body"),
     ]
+
+
+class ShowcasePage(Page, SocialMediaMixin, CrossPageMixin):
```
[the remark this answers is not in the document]
William-Blackie: Sure thing!

wagtailio/core/models/pages.py
William-Blackie: Sure thing.

wagtailio/core/models/pages.py
William-Blackie: I've wrapped this in a `MultiFieldPanel` panel, although this site is still 4.2 so we cannot show icons until 5x :).

wagtailio/core/models/pages.py
William-Blackie: It can be, I guess this is just a preference of mine to keep the FE logic simple. Also, this site isn't a multi site so I think the query is the same regardless?

wagtailio/core/models/pages.py
William-Blackie: Sure thing, this is neat.

wagtailio/core/models/pages.py
William-Blackie: I've not really been given any brief of how long this should be, I tend to prefer leaving this to editor control unless there's a reason to lock it down?

wagtailio/project_styleguide/templates/patterns/pages/showcase_page/showcase_page_fragment.html
William-Blackie: Yeah, that's a good spot. let's enforce it and keep it consistent in the UI.

wagtailio/core/models/pages.py
William-Blackie: I've made this change anyway, it could be someday.

wagtailio/core/models/pages.py
zerolab: if it is kept to `p, a, ul, li, link, document` that should cover most uses without needing special styling or handling for something like headings

wagtailio/showcase/migrations/0003_add_listing_meta_description.py
```diff
@@ -0,0 +1,26 @@
+# Generated by Django 3.2.18 on 2024-01-03 15:37
```
zerolab, line 1: nitpick: can this be part of the initial migration?

wagtailio/project_styleguide/templates/patterns/pages/showcase_page/showcase_page.html
```diff
@@ -0,0 +1,53 @@
...
+{% block title %}Wagtail CMS - {% firstof page.seo_title page.title %}{% endblock %}
+{% block title_suffix %}{% endblock %}
+
+{% block content %}
+    <div class="showcase-page__hero showcase-page__content">
+        <h1 class="showcase-page__title">{{ page.title }}</h1>
+        {% if page.introduction %}<div class="showcase-page__intro">{{ page.introduction }}</div>{% endif %}
+    </div>
+    <div class="showcase-page__content">
+        <form
```
zerolab, line 20: question: what happens if people have no JS?

wagtailio/project_styleguide/templates/patterns/pages/showcase_page/showcase_page_fragment.html
```diff
@@ -0,0 +1,86 @@
...
+                    <p class="showcase-card__description">{{ item.description }}</p>
+                    <a class="showcase-card__link" href="{% if item.external_link %}{{ item.external_link }}{% else %}{% pageurl item.internal_link %}{% endif %}">{{ item.link_text }} <svg class="showcase-card__svg" width="15" height="16" aria-hidden="true">
+                        <use xlink:href="#{% if self.external_link %}open-link{% else %}arrow{% endif %}"></use>
+                    </svg></a>
+                </div>
+            </div>
+        {% endfor %}
+    </div>
+    <div>
+        <div class="pagination pagination--showcase">
```
zerolab, line 60: question: why not hook the pagination to HTMX too?

wagtailio/showcase/models.py
```diff
@@ -0,0 +1,176 @@
...
+from wagtailio.utils.models import CrossPageMixin, SocialMediaMixin
+
+
+class ShowcasePage(SocialMediaMixin, CrossPageMixin, Page):
+    template = "patterns/pages/showcase_page/showcase_page.html"
+    htmx_template = "patterns/pages/showcase_page/showcase_page_fragment.html"
+
+    parent_page_types = ["core.HomePage"]
+
+    introduction = models.TextField(
```
zerolab, line 22: I still think this should be a `RichTextField` with `p, a, ul, ol, link, document-link`

wagtailio/showcase/models.py
```diff
@@ -0,0 +1,176 @@
...
+
+    promote_panels = (
+        Page.promote_panels + SocialMediaMixin.panels + CrossPageMixin.panels
+    )
+
+    search_fields = Page.search_fields + [
+        index.SearchField("introduction"),
+    ]
+
+    def _filter_used_sectors(self):
```
zerolab, line 52: nitpick: this reads like an action to me. How about something like `_get_used_sectors()` and drop the "All" value (or better said, shift "All" to `get_context`)

wagtailio/showcase/migrations/0003_add_listing_meta_description.py
William-Blackie: That's a fair ask, I generally only do this once for the initial review, and subsequently add new migrations so the reviewer does not have to handle their DB state as much.

wagtailio/project_styleguide/templates/patterns/pages/showcase_page/showcase_page_fragment.html
William-Blackie: Yep, can do!

wagtailio/project_styleguide/templates/patterns/pages/showcase_page/showcase_page.html
William-Blackie: Fair question, as of late, I'm less and less concerned with non-JS, especially due to being deproritised by GDS. Although, the only real annoyance this requires is a small amount of pop-out from hidding a submit button, which we can do.

wagtailio/showcase/models.py
William-Blackie: Fair, sorry, I didn't mean to ignore this!.

wagtailio/showcase/models.py
William-Blackie: Done, I've also extended the method slightly due to handling the FakeQuerySet in the preview.

wagtailio/core/models/pages.py
William-Blackie: You got it :)

wagtailio/core/migrations/0061_create_initial_sectors.py
William-Blackie: I've added a backwards method. The answer is the previous migration removes this model so I guess I just thought to remove migrate backwards is the same when you include the scope of the PR.

wagtailio/core/models/pages.py
William-Blackie: Done.

wagtailio/project_styleguide/templates/patterns/pages/showcase_page/showcase_page.html
William-Blackie: Done

wagtailio/project_styleguide/templates/patterns/pages/showcase_page/showcase_page_fragment.html
```diff
@@ -0,0 +1,99 @@
...
+                </div>
+            </div>
+        {% endfor %}
+    </div>
+    <div>
+        <div
+            hx-boost="true"
+            class="pagination pagination--showcase">
+            {% if showcase_items.has_previous %}
+                <a hx-get="{% pageurl page %}"
```
nanuxbe, line 64: It feels counter-intuitive that the `hx-get` and `href` are different. Is there a reason why they are different? Since `hx-push-url` is set to `True`, it feels like both urls should be callable. Is there a difference between the 2?

wagtailio/project_styleguide/templates/patterns/pages/showcase_page/showcase_page_fragment.html
```diff
@@ -0,0 +1,99 @@
...
+                Page {{ showcase_items.number }} of {{ showcase_items.paginator.num_pages }}
+            </span>
+
+            {% if showcase_items.has_next %}
+                <a
+                    hx-trigger="click"
+                    hx-target="#showcase_results"
+                    hx-push-url="true"
+                    hx-swap="outerHTML"
+                    hx-swap="innerHTML"
```
nanuxbe, line 88: You have `hx-swap` twice and no `hx-get` here

nanuxbe (changes requested):
Here are a couple issues I noticed. I mostly looked at the HTMX code.

wagtailio/project_styleguide/templates/patterns/pages/showcase_page/showcase_page_fragment.html
William-Blackie: Thanks for the review @nanuxbe! This is a good spot, really we don't even need the hx-get as hx-boost makes these links compatible and the same thing.

wagtailio/project_styleguide/templates/patterns/pages/showcase_page/showcase_page_fragment.html
William-Blackie: Good spot for the duplication, we don't need hx-get as above it we have hx-boost! :)

William-Blackie:
Update: On staging.

vossisboss:
Am reviewing the content and such. Still reviewing but wanted to flag these system warnings that appeared when I ran the branch locally.

```
?: (templates.E003) 'wagtailcore_tags' is used for multiple template tag modules: 'wagtail.templatetags.wagtailcore_tags', 'wagtailio.project_styleguide.templatetags.wagtailcore_tags'
?: (templates.E003) 'wagtailembeds_tags' is used for multiple template tag modules: 'wagtail.embeds.templatetags.wagtailembeds_tags', 'wagtailio.project_styleguide.templatetags.wagtailembeds_tags'
?: (templates.E003) 'wagtailimages_tags' is used for multiple template tag modules: 'wagtail.images.templatetags.wagtailimages_tags', 'wagtailio.project_styleguide.templatetags.wagtailimages_tags'
?: (templates.E003) 'wagtailsearchpromotions_tags' is used for multiple template tag modules: 'wagtail.contrib.search_promotions.templatetags.wagtailsearchpromotions_tags', 'wagtailio.project_styleguide.templatetags.wagtailsearchpromotions_tags'
```

vossisboss:
@William-Blackie Can you confirm this branch is on staging? I'm not seeing the Showcase page child page option on staging but it is on my local copy of the branch. I'll proceed with reviewing the content and such based on my local copy.

zerolab:
@vossisboss those warnings relate to django-pattern-library and Django 4.2+

vossisboss:
@William-Blackie Found a few things we should address before pushing to production. Note that I ran my tests locally and not in staging.

## The logo display needs some adjustment. 

In Ben's design, there is a solid background behind the logos to avoid the issue of the logo clashing with the image. It looks like there is a background there, but the transparency might be off? Either way, it needs to be addressed.
Screenshot 2024-02-09 at 4 50 07 PM

## Move the pagination below the CTA

When there is more than one page, it would be ideal if the pagination was below the CTA so that it doesn't get shoved down to the bottom of the page when there are multiple pages.
Screenshot 2024-02-09 at 5 10 31 PM


## Adjust text padding in the mobile display

The text padding in the mobile version seems a bit sparse. It would be nice if there was a bit more space between the words and the edge of the screen.
Screenshot 2024-02-09 at 5 09 57 PM


I'm also not sure why the meta text is at the bottom of the page. It would be more useful to have it beneath the intro text. But to be fair to you, it is in Ben's design. So I'll run that one by Ian and Ben.

Thanks so much for putting this together. This will be a real useful page type once we have things spruced up!

benenright:
Yes, the logos should have a solid white bg and be a fixed size with the logo placed centrally with like 10px padding around it: 
image

----

There should also be a very subtle drop shadow on the logo bg:
image

----

Sorry for not supplying a mobile design but yes the left/right/bottom padding on the text part of cards is too tight. Make it all 25px.

----

This:
image
Was supposed to be this footnote:
image
...but with the paging it feels wierd (do we need it?)

----

Why is the 'next' button so wide in the pagination? Should be more like this:
image


Cheers
Ben

vossisboss:
@benenright FYI, I think the Next button is large in the example I shared because there is no "Previous" button on the first page.

@William-Blackie Ben and I discussed the meta text and where it should be placed. It would be great to have the option to use it closer to the top, right below the line in the prototype that says "Sites and apps made with Wagtail." We would like to have the option to add a bit of descriptive text if needed.

William-Blackie:
> @benenright FYI, I think the Next button is large in the example I shared because there is no "Previous" button on the first page.
> 
> @William-Blackie Ben and I discussed the meta text and where it should be placed. It would be great to have the option to use it closer to the top, right below the line in the prototype that says "Sites and apps made with Wagtail." We would like to have the option to add a bit of descriptive text if needed.

Hi @benenright and @vossisboss, 

Thanks for the feedback! I'll split my responses a little for readability, anything I don't reference I'll make the changes shortly.

For the logo, I assumed that the images themselves would have backgrounds for two reasons; the first being having a solid background colour could be an accessibility issue, i.e. contrast but I guess that should be a minor worry as most are designed against both light and dark for this reason. The second being to get a consistent sizing it's much easier to supply an image that's to be resized with a background to keep an aspect ratio than allow editors to supply this. But, we can get around this by not scaling the images up, and instead rely on editors to do this. I guess this is mostly force of habit from client work!

I see your points about the pagination, the request came from Ian B as a must, so I that would need to be queried by him. It wasn't in the designs so I repurposed the existing pagination. I'll spend some more time on styling if this is still needed?

For the meta text, "List curated by the Wagtail core team", your request is a little confusing. Are you instead asking for a richtext field beneath the introduction? As this is already a richtext field we could already have this functionality unless you have a different idea in mind. 

Let me know what you think; thanks - Will.

vossisboss:
@Tijani-Dia I went ahead and made the field fixes that we needed. I also removed the unnecessary meta field and moved the CTA so that it is above the pagination when there is more than one page displayed.

When I was testing out the pagination by adding more than six items though, I discovered a potential bug. When you click "Next" to get to Page 2 and then "Previous" to get back to Page 1, everything looks fine. But if you repeat going back and forth navigating the pages with the "Next" and "Previous" buttons, something weird happens and the page starts to look like this:

Screenshot 2024-03-11 at 9 37 25 PM

Screenshot 2024-03-11 at 9 37 35 PM

It's like the page is repeating itself somehow. If you could add more than six showcase items to the page and see if you can get it to break in the same way, it would be nice to know if this is really a bug.

We can do a Slack huddle tomorrow if you need me to show you what I mean.

vossisboss:
@Tijani-Dia  and I did some further investigation of this bug and discovered that it is specific to the Firefox browser. We couldn't replicate it in other browsers. Tidiane suspects that is related to the HTMX rendering. Since this bug is relatively hard for a user to trigger, we're going to make a note of it and proceed to testing this branch in staging.

vossisboss:
@thibaudcolas confirmed the bug in Firefox and found an additional bug in Chrome related to HTMX. Per @thibaudcolas's recommendation, I'm going to remove the HTMX since the vanilla HTML seems to work just fine.

vossisboss:
@zerolab and/or @thibaudcolas This branch is ready for review if either of you are available to provide a second pair of eyes on it. Please note that the HTMX was removed due to the bug I reported above. A test showcase page is available on the staging site as well if you would like to look at it. I also haven't rebased the branch yet. I'd like to know if any other big changes are needed before I do that.

wagtailio/static/js/main.js
```diff
@@ -34,6 +35,8 @@ document.addEventListener('DOMContentLoaded', () => {
     initComponent(MobileMenu);
     initComponent(MobileSubMenu);
     initComponent(DesktopSubMenu);
+    initComponent(ProgressiveEnhancement);
```
thibaudcolas, line 38: I don’t see this used anywhere currently, so could be removed?

thibaudcolas (commented):
Backend code review – switching over to the front-end after.

One extra point to check: Migrations – not sure if we have a policy to squash all the schema changes into the initial migration for something like this? Noting this is on staging already though.

wagtailio/project_styleguide/templates/patterns/pages/showcase_page/showcase_page.yaml
```diff
@@ -0,0 +1,18 @@
+context:
```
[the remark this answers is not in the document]
vossisboss: Are the yaml files primarily for the pattern library? This is one of those things I haven't quite figured out how to get working. The couple of times I did encounter the pattern library, it did seem fairly buggy to me.

Issue #322, opened 2022-11-10
font-size definitions shouldn’t be in absolute units


thibaudcolas:
There are a large number of `font-size` definitions set in `px` values across the project. This is problematic for users of browser-level text zooming. All those definitions should be replaced with equivalent `rem` values.

Aarthy153:
Hi, I would like to work on this issue. Can I get assigned?

zerolab:
@Aarthy153 we do not assign issues to non-core folk. Add a note that you are working on the issue, then open a PR

kituuu:
@thibaudcolas @zerolab, can I work on it if the issue is still not fixed?

thibaudcolas:
Anyone can work on anything, please read the comment right above yours.

PR #448, opened 2024-01-03
style: changed font-size units from px to rem

kituuu:
#322

### DIFF
--- a/wagtailio/newsletter/templates/newsletter/newsletter_page.html
+++ b/wagtailio/newsletter/templates/newsletter/newsletter_page.html
@@ -354,20 +354,20 @@
                 word-break: normal;
             }
 
-            h1 {font-size: 40px; font-weight:700; margin-bottom: 20px;}
-            h2 {font-size: 36px; font-weight:700; margin-bottom: 20px;}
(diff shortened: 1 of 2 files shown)


thibaudcolas:
@kituuu thank you for giving this a go. Could you take a moment to describe how you’ve approached this task, and how you’ve tested your work?

kituuu:
> @kituuu thank you for giving this a go. Could you take a moment to describe how you’ve approached this task, and how you’ve tested your work?

Hey, so basically I used the calc method available in css to convert all the px sizes to rem. Like for example 1px is 0.0625rem. Suppose my font size is 20px, so replaced it with calc(20*0.0625rem) , 20*0.0625 is 1.5 so it evaluates to 1.5rem. 

I have tested some of the tags by making a some dummy pages, and then checked the rendered UI for px units and rem units side by side with the help of developer tools. 

As there are so many components/sections I can't check all of them. But the once which I tested are working as intended. As the logic is same everywhere, so I suppose that it will work with all components/sections.

thibaudcolas (approved):
👍 this is working as expected for the newsletter pages and the 500 page… but not for any of the other component, where we use dynamic font sizes based on media queries on the root element.

Issue #454, opened 2024-01-29
WHCM styles improvements


thibaudcolas:
Based on the audit from @shyusu4, see [Wagtail.org | Accessibility audit spreadsheet](https://docs.google.com/spreadsheets/d/1kWb6f9YlT-UA2TC4ioAH40Y5T18hd_Hbt_ncTamabW4/edit). WHCM fixes:

## Issues

- [x] Close icon in the search modal isn’t visible.
- [x] Menu toggle isn't visible.
- [x] Copy icon in "pip install Wagtail" button disappears when clicking on it.
- [x] (not in audit) SVG icons in buttons now match the button text color
- [x] Social media icons are not using the correct color (linktext)
- [x] All social media, “sponsor wagtail”, and “contribute code” icons aren't visible.
- [ ] #463
- [x] Arrow icon within CTA button isn't visible when hovered.
- [x] Arrow icon within CTA link in sub-navigation panel isn’t visible.
- [x] (Blog posts) Code blocks should have a clear separation from regular content.

## Working on this

Assigning this to @shyusu4 as part of her Outreachy project

thibaudcolas:
Closing with last remaining item at #463.

PR #455, opened 2024-01-29
WHCM styles improvements

shyusu4:
#454
## Description
This pull request addresses styling issues related to windows high contrast mode. The following issues have been fixed:

- **CTA button:** 
    - Arrow icon isn't visible when hovered.
    - Inconsistent transition speeds between the title and subtitle.
- **Header navigation:** Arrow icon within CTA link in sub-navigation panel isn’t visible.
- **Code snippet block:** Code blocks should have a clear separation from regular content.
- **Footer:** 
  - All social media, “sponsor Wagtail”, and “contribute code” icons aren't visible. 
  - Social media icons are not using the correct color (link text). 

_Note_ - Dark/light mode theme toggle was meant to not be displayed in HCM but used `prefers-contrast` instead of `forced-colors` active.

<details>

  | | | |
  | --- | --- | --- |
  | Screenshot (65) | Screenshot (63) | Screenshot (64) |
  | Screenshot (68) | Screenshot (69) | Screenshot (70) |
  | Screenshot (72) | Screenshot (73) | Screenshot (74) |
  | Screenshot (77) | Screenshot (75) | Screenshot (76) |

</details>

### DIFF
--- a/wagtailio/static/sass/components/_cta.scss
+++ b/wagtailio/static/sass/components/_cta.scss
@@ -28,7 +28,11 @@
     > svg {
         transition: transform 0.3s;
         margin: 0 0 0 auto;
-        flex-shrink: 0
+        flex-shrink: 0;
+
+        @media (forced-colors: active) {
+            fill: currentColor;
+        }
     }
 
     &:hover {
@@ -42,13 +46,16 @@
         > svg {
             fill: $color--white;
             transform: translateX(10px) scale(1.1);
+
+            @media (forced-colors: active) {
+                fill: currentColor;
+            }
         }
     }
 
     &__title {
         color: var(--color--interaction);
         margin-bottom: 10px;
-        transition: color $transition;
 
         @include media-query(medium) {
             margin-bottom: 5px;
--- a/wagtailio/static/sass/components/_nav-cta.scss
+++ b/wagtailio/static/sass/components/_nav-cta.scss
@@ -93,8 +93,8 @@
         position: absolute;
         bottom: 30px;
         right: 20px;
-        stroke: $color--white;
-        stroke-width: 1px;
+        stroke: currentColor;
+        stroke-width: 4px;
         fill: transparent;
         width: 60px;
         height: 60px;
--- a/wagtailio/static/sass/components/_nav-cta.scss
+++ b/wagtailio/static/sass/components/_nav-cta.scss
@@ -93,7 +93,7 @@
         position: absolute;
         bottom: 30px;
         right: 20px;
-        stroke: currentColor;
+        stroke: $color--white;
         stroke-width: 4px;
         fill: transparent;
         width: 60px;
@@ -103,6 +103,10 @@
         @media only screen and (min-width: $header-desktop) {
             bottom: 50px;
             right: 60px;
+        }
+
+        @media (forced-colors: active) {
+            stroke: currentColor;
         }
     }
 }
--- a/wagtailio/static/sass/layout/_footer.scss
+++ b/wagtailio/static/sass/layout/_footer.scss
@@ -5,20 +5,6 @@
         background-color: var(--color--highlight);
         padding: 60px 0;
         color: $color--white;
-
-        @media (forced-colors: active) {
-            a:link, a:visited {
-                svg {
-                  fill: currentColor;
-                  fill: LinkText;
-                }
-            }
-            a:hover {
-                svg {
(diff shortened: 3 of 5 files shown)


wagtailio/static/sass/components/_cta.scss
```diff
@@ -28,7 +28,11 @@
     > svg {
         transition: transform 0.3s;
         margin: 0 0 0 auto;
-        flex-shrink: 0
+        flex-shrink: 0;
+
+        @media (forced-colors: active) {
```
albinazs, line 33: Great use of `currentColor`!

wagtailio/static/sass/components/_cta.scss
```diff
@@ -42,13 +46,16 @@
         > svg {
             fill: $color--white;
             transform: translateX(10px) scale(1.1);
+
```
albinazs, line 49: I think repeating this rule on hover is unnecessary though

wagtailio/static/sass/components/_nav-cta.scss
```diff
@@ -93,8 +93,8 @@
         position: absolute;
         bottom: 30px;
         right: 20px;
-        stroke: $color--white;
-        stroke-width: 1px;
+        stroke: currentColor;
+        stroke-width: 4px;
```
albinazs, line 97: Here we're affecting designs in a non-contrast mode, we need to add media query to target forced-colors only

wagtailio/static/sass/layout/_footer.scss
```diff
@@ -176,6 +176,9 @@
             transition: fill  $transition;
         }
 
+        @media (forced-colors: active) {
+            fill: currentColor;
```
albinazs, line 180: nice touch to make icons more prominent for users relying on forced-colors!

albinazs:
thank you, @shyusu4! great stuff!:)

wagtailio/static/sass/components/_cta.scss
```diff
@@ -42,13 +46,16 @@
...
+            @media (forced-colors: active) {
+                fill: currentColor;
+            }
         }
     }
 
     &__title {
         color: var(--color--interaction);
         margin-bottom: 10px;
-        transition: color $transition;
```
albinazs, line 51: not sure why this has been removed

shyusu4: This was removed as part of another minor issue mentioned in the audit (not whcm related). The CTA button currently has inconsistent transition speeds between the title and subtitle in light mode.

wagtailio/static/sass/components/_cta.scss
shyusu4: Because the arrow icon is white when hovered, it practically disappears for contrast themes with light backgrounds. That's why I chose to repeat the rule here.

albinazs: good point!

wagtailio/static/sass/components/_cta.scss
albinazs: Ok, in this case, can you please add this issue to the PR description

wagtailio/static/sass/components/_nav-cta.scss
Scotchester: The `stroke: currentColor` does not seem to have any effect on the regular rendering. In my opinion the `stroke-width` increase would be an improvement to the base design, it's not my place to decide that, so I agree with Albina's suggestion to move these rules into a media query.

Scotchester:
Good stuff, @shyusu4! Just the one thing where we need the media query that Albina noticed before we can approve.

wagtailio/static/sass/components/_nav-cta.scss
shyusu4: Thanks @Scotchester @albinazs - I've moved the use of `currentColor` into a media query, but left `stroke-width` because this is part of another minor issue in the audit. I think it would be better to keep the increased width so that the nav CTA is more distinguishable as a link.

Scotchester: Ah, gotcha! If we are considering the thin stroke an accessibility issue as well, then I am happy to have the change :)

Scotchester (approved):

thibaudcolas:
🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀

Issue #457, opened 2024-01-31
Decorative shapes overlap section subheadings


shyusu4:
## Issue
Decorative shapes are currently overlapping with section subheadings.

<img alt="Decorative shape overlapping section subheading in dark mode of wagtail.org" width="250" src="https://github.com/wagtail/wagtail.org/assets/106587342/b1c4ee8f-b329-4c72-aa41-3dcafe6ea6fc">
<img alt="Decorative shape overlapping section subheading on website in light mode of wagtail.org" width="250"src="https://github.com/wagtail/wagtail.org/assets/106587342/4085aad1-827a-469d-ab81-757d9d6aef7f">

**Steps to reproduce**
Resize the viewport to 768px or less.

**Suggested fix**
Make sure icons are always behind text elements in z-index layering or/and add a solid background according to theme behind the text. 

**Reference**
[Wagtail.org | Accessibility audit spreadsheet](https://docs.google.com/spreadsheets/d/1kWb6f9YlT-UA2TC4ioAH40Y5T18hd_Hbt_ncTamabW4/edit?usp=sharing)

Abhii67534:
Please assign this to me :)

thibaudcolas:
There’s been some improvements in #458 but I think there’s more work to do on this.

PAVAN507-STAR:
@thibaudcolas I would love take on this issue sir.If possible can u guide me on what more changes to add.

thibaudcolas:
@PAVAN507-STAR I think the suggested fix above is good?

> 1. Make sure icons are always behind text elements in z-index layering
> 2. add a solid background according to theme behind the text.

They both sound like valuable improvements.

PAVAN507-STAR:
@thibaudcolas Okay sir will do that in 2-3 days  if that's fine as rn i have clg exams going on.

oz456:
Hey @thibaudcolas can you review my PR #499 resolving this issue.

thibaudcolas:
Thank you @oz456, this definitely helps. This doesn’t change the second point from the suggested fix:

> 2. add a solid background according to theme behind the text.

oz456:
As I see, on different themes the text above the background icon looks like this: 
| Dark Theme | Light Theme |
|--------|--------|
[...]

PR #458, opened 2024-02-05
Headings and forms accessibility improvements

shyusu4:
This PR addresses accessibility issues related to heading levels and forms:

- Fixed instances (in hero, headline, and icon-link blocks) where headings were used consecutively after one another.
- Changed menu_get_started to use an `h2`
- #457
- #459
- Currently screen reader announces sign up forms on the site with "Enter your email address" repeated 3 times. Fixed by changing base_form to use `aria-label="Email"` instead of `<label>`
- Changes the use of `fieldset` in search-form to a `div` (fieldset is used to group multiple form fields together, but there's only one here)

<details>

  | before | after |
  | --- | --- |
  |Screenshot (109)|Screenshot (105)|
Screenshot (110)|Screenshot (106)|

</details>

### DIFF
--- a/wagtailio/project_styleguide/templates/patterns/includes/search-form.html
+++ b/wagtailio/project_styleguide/templates/patterns/includes/search-form.html
@@ -1,10 +1,10 @@
 <form action="{% url 'search' %}" method="get" role="search" class="search {% if classes %}{{ classes }}{% endif %}" {% if data_attrs %}{{ data_attrs }}{% endif %} {% if form_id %}id="{{ form_id }}"{% endif %}>
-    <fieldset class="search__wrap">
+    <div class="search__wrap">
         <label class="u-sr-only" for="{{ input_id }}">Enter a search term</label>
         <input class="search__input" id="{{ input_id }}" type="search" name="query"{% if search_query %} value="{{ search_query }}"{% endif %} placeholder="Search...">
         <button type="submit" class="button button--secondary button--form search__button" aria-label="Search the site content">
             <span class="button__text">Search</span>
             <svg class="arrow" aria-hidden="true"><use xlink:href="#arrow"></use></svg>
         </button>
-    </fieldset>
+    </div>
 </form>
--- a/wagtailio/static/sass/components/_sign-up-form.scss
+++ b/wagtailio/static/sass/components/_sign-up-form.scss
@@ -3,7 +3,7 @@
     @include sf-spacing(2);
     grid-column: 2 / span 2;
     overflow: hidden; // prevent scrollbars from svg size
-
+    
     @include media-query(medium) {
       grid-column: 2 / span 3;
     }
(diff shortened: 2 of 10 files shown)


Scotchester:
> Fixed instances (in hero, headline, and icon-link blocks) where headings were used consecutively after one another.

I looked for an issue in the audit spreadsheet that this addresses but couldn't easily find one. Could you point it out for me?

I'm a bit concerned about the addition of `hgroup` elements. They're probably harmless in the short term, but [according to MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/hgroup), they have been removed from the HTML spec and are discouraged from being used. If a wrapper is helpful for applying the `role="group" aria-roledescription="heading group"` attributes, a `div` is probably best in this case.

wagtailio/static/sass/components/_sign-up-form.scss
```diff
@@ -25,19 +25,22 @@
...
         }
 
         > svg {
             position: relative;
-            inset: -100px 0 auto auto;
             width: 170px;
             height: 170px;
+
+            @include media-query(medium) {
+                inset: -100px 0 auto auto;
```
Morsey187, line 42: Nickpick: This is a huge improvement over the current styling, however it's still possble to have the text overlap at some viewports, which given the bright colour of the icon here makes the text really difficuly to read.

<details><summary>Example</summary>

Screenshot 2024-02-06 at 17 39 07

</details>

If you altered the DOM a little, you could then constrain the text to a division element and have both the text and icon be responsive to their respective containers, using `display: flex`.


<details><summary>Example using flex</summary>
You could then use `flex-direction: column` to move the icon below the form at mobile. We'd also have to wouldn't have to offset the image using CSS which can be tricky to manage. 

Screenshot 2024-02-06 at 17 44 03

</details>

Morsey187 (commented):
This looks great, I haven't looked so much into the use of aria tags and  hgroup as I noticed Scot's comment and certain things were discussed already on Zoom so I think Thibaud is better placed to review that aspect. 

Instead I've checked out the branch and tested locally, the SCSS looks fine to me I've just left some nickpicks/suggestions around text overlapping.

wagtailio/project_styleguide/templates/patterns/includes/search-form.html
```diff
@@ -1,10 +1,10 @@
 <form action="{% url 'search' %}" method="get" role="search" class="search {% if classes %}{{ classes }}{% endif %}" {% if data_attrs %}{{ data_attrs }}{% endif %} {% if form_id %}id="{{ form_id }}"{% endif %}>
-    <fieldset class="search__wrap">
+    <div class="search__wrap">
```
thibaudcolas, line 2: Looking good!

thibaudcolas (commented):
There are a few things that need changing, and I think more work needed on the text overlap, but this goes in the right direction.

thibaudcolas (approved):
Merging without the `role` and `aria-roledescription` usage!

PR #462, opened 2024-02-19
Updating the image settings to allow avif and webp image formats.

vossisboss:
Due to the default Wagtail behavior of converting avif and webp files to PNGs, we need these settings to allow for these newer and greener file formats to be used on Wagtail.org.

I have decided that will not support GIFs as they tend to be large, unwieldy files and they are a potential source of chaos if animated versions are allowed to be uploaded.


zerolab (approved):

zerolab:
pending linting fixes, of course

vossisboss:
@zerolab You're online late. Just committed the linting fix. Checks running.

PR #464, opened 2024-02-26
Apply richtext styles to markdown blocks and improve spacing of markdown/code blocks

kbayliss:
This MR will:

1. Apply the same rich text styles applied to `RichTextBlock` to `MarkDownBlock`, which fixes display issues of `ul` and makes other elements display consistently within `MarkDownBlock`s.
2. Standardise the spacing of code blocks within `MarkDownBlock` and `CodeBlock`

### Screenshots

Both "Before" and "After" screenshots below are taken using the same StreamField which uses two approaches to achieve the same content:

Streamfield

#### Before
Before

#### After
After

### DIFF
--- a/wagtailio/project_styleguide/templates/patterns/components/streamfields/code_block/code_block.html
+++ b/wagtailio/project_styleguide/templates/patterns/components/streamfields/code_block/code_block.html
@@ -1,6 +1,6 @@
 {% load wagtailcore_tags %}
 <div class="grid">
-    <div class="grid-align">
+    <div class="code-block">
         {{ code }}
     </div>
 </div>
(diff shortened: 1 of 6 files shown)


wagtailio/project_styleguide/templates/patterns/components/streamfields/code_block/code_block.html
```diff
@@ -1,6 +1,6 @@
 {% load wagtailcore_tags %}
 <div class="grid">
-    <div class="grid-align">
```
kbayliss, line 3: Renamed because this is the only place `.grid-align` was used, so `.code-block` makes more sense

vossisboss (approved):
This looks good @kbayliss. I tested this with three previous blog posts that contain code blocks as well and they all look good. I'm not aware of any other ways we use the `code-block` styling outside of the blog, so this is seems good to go to me.

PR #465, opened 2024-02-27
Remove surplus margin from richtext elements

kbayliss:
This MR removes the bottom margin from the last element (`:last-child`) of rich text elements so that we rely on the parent block's spacing rather than that _plus_ whatever margin exists on the last element.

The example below removes the bottom margin from the `ul`.

**Before:**

Before

**After:**

After


thibaudcolas (approved):
ty!

Issue #412, opened 2023-10-02
Reduce size of large headings


benenright:
To better accommodate longer headlines and feel a bit less shouty, can we reduce the font-size and line-height of the large headlines to 4rem and 4.4rem on desktop:

![image](https://github.com/wagtail/wagtail.org/assets/4907819/b4ceae28-c301-49b4-b388-9a690ce65e54)

<img width="1390" alt="image" src="https://github.com/wagtail/wagtail.org/assets/4907819/0ce1d51f-9169-45ec-8a88-055896379b8a">

<img width="1368" alt="image" src="https://github.com/wagtail/wagtail.org/assets/4907819/d997072f-36d5-4639-84b4-e81ea5a8edfb">

benenright:
Similarly, reduce H2's within the blog page body on desktop.

![image](https://github.com/wagtail/wagtail.org/assets/4907819/23da6582-6855-4d87-a053-7a5ec1abed77)

<img width="866" alt="image" src="https://github.com/wagtail/wagtail.org/assets/4907819/aaef4ce1-7153-4246-98ba-977e7b238ec8">

oly-baby:
hello, can i work on this

thibaudcolas:
@oly-baby 👋 thanks for your interest in this! You can go ahead. No need to ask for permission, everything on our issue tracker is there for the taking unless someone else has already been working on it.

Please take a moment to understand the task at hand, make sure you can reproduce the current state of things on your copy of the project, and then implement the desired changes as per what the issue describes.

oly-baby:
Alright 👍

OsafAliSayed:
Is this issue closed or still needs working?

benenright:
Hi @chris-lawton 

Here are the values I'd like to update to (just on desktop for now):

### h1, .heading-one 
![image](https://github.com/wagtail/wagtail.org/assets/4907819/01327b0a-f32d-4afd-a9c4-ec9a2f0a4b92)

### .intro-big
![image](https://github.com/wagtail/wagtail.org/assets/4907819/ac984f70-27e6-4f61-bd72-1427be478139)

### h2 
![image](https://github.com/wagtail/wagtail.org/assets/4907819/1b982eb5-99f4-42ee-a676-709e08808eff)

benenright:
Sorry, i got the H2 one wrong above. Should also be:
[...]

PR #472, opened 2024-06-10
Update typography sizes and reduce spacing

chris-lawton:
Issue: https://github.com/wagtail/wagtail.org/issues/412

Reduces the sizes of the base h1 and h2 tags above the large breakpoint and does the same for the `.intro-big` class.

Also reduce the bottom margin on all base heading tags.

Updated to units provided by Ben here: https://github.com/wagtail/wagtail.org/issues/412#issuecomment-2147570850


thibaudcolas (approved):
Thank you @chris-lawton, all looking good!

PR #473, opened 2024-06-21
Wagtail 6.1 and Django 5.0 upgrade

Stormheg:
Think it would be nice if wagtail.org would run the latest Wagtail and Django. Unless we have reason to stay on long term support versions?

I don't have an up-to-date copy of the database to test if nothing breaks. Hope someone is able to help out with this.

Especially the dependency on `wagtailaltgenerator` is concerning to me as it has seen no updates in about 4 years. I think we should remove it and add alt text manually.


allcaps (approved):
LGTM

KalobTaulien (commented):
I haven’t tested this locally, yet. I might not get a chance to spin this up with real..ish data and media. But at a glance this looks good to me. 👍

zerolab (approved):
Took it for a spin locally. LGTM

we can drop `wagtailaltgenerator` - it was used for various demos in the past, but it is not something we actively use at the moment as far as I know. /cc @vossisboss

Stormheg:
Great, thanks all! I'll add a commit to drop `wagtailaltgenerator` 👍

vossisboss:
@Stormheg Thanks so much for this! Sorry I didn't see this sooner. @cnk and I caught that the Poetry version needed to be updated in the CI. But otherwise I think we're ready to merge this! 🎉

PR #474, opened 2024-06-23
Update to use Node 20 and various frontend dependency updates

Stormheg:


thibaudcolas (approved):
All seems to be working well, except the `npm run start:reload` due to the webpack-dev-server upgrade. Will fix that and merge!

PR #475, opened 2024-07-22
Upgrade postgres to 16

RealOrangeOne:

PR #477, opened 2024-10-16
Update All The Things October 2024

vossisboss:
Started a branch with @cnk to bump to Wagtail 6.2


Stormheg:
@vossisboss let me know if you need reviews or help 😄

vossisboss:
Hey @Stormheg, you mind having a look at the linter error that is popping up? I'm not sure why flake8 is flagging `needs_sponsorship` in the models.py file of the roadmap app. Seems odd to me.

laymonage:
Huh, looks like I made a mistake in 2ecbe815cd14e748cba7add48efd7d853ba99704 where I should've also removed the field definition from the model:
https://github.com/wagtail/wagtail.org/blob/ebf3f4c00adf368626e450660dde16d7c6e863b9/wagtailio/roadmap/models.py#L77

It should be OK to remove it, its functionality has been replaced by the `needs_sponsorship` cached property that looks at the labels from GitHub rather than it being a model field:

https://github.com/wagtail/wagtail.org/blob/ebf3f4c00adf368626e450660dde16d7c6e863b9/wagtailio/roadmap/models.py#L132-L134

That property overwrites the model field in the same namespace, hence flake8 tripped. I guess since you updated the linter versions in a commit in this PR, that means it's a new rule added by flake8 in newer versions. Could you try removing that `BooleanField` and see if it works @vossisboss? Thank you for working on this!

vossisboss:
The final bits still needs to deploy to staging, but overall this is ready for a review. Could I interest either @Stormheg or @cnk or @zerolab in reviewing this upgrade before we go live.

Also FYI @RealOrangeOne I added the Postgres upgrade here before I saw your PR.

Dockerfile
```diff
@@ -11,7 +11,7 @@ RUN npm run build:prod
 
 
 # Build Python app - this stage is a common base for the prod and dev stages
-FROM python:3.11-bullseye AS backend
+FROM python:3.12-bullseye AS backend
```
zerolab, line 14:
```suggestion
FROM python:3.12-bookworm AS backend
```

ref: https://endoflife.date/debian

wagtailio/roadmap/models.py
```diff
@@ -74,7 +74,6 @@ def get_context(self, request, *args, **kwargs):
 class MilestoneItem(Orderable):
     NEEDS_SPONSORSHIP_LABEL = "needs sponsorship"
 
-    needs_sponsorship = models.BooleanField(default=False, editable=False)
     sponsorship_url = models.URLField(
```
zerolab, line 77: does this not need a migration?

laymonage: No it does not, see https://github.com/wagtail/wagtail.org/pull/477#issuecomment-2435942999

zerolab: ah, I missed your comment. Thank you

Dockerfile
Stormheg: Good one Dan!

.github/workflows/ci.yml
```diff
@@ -15,10 +15,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - name: Set up Python 3.11
+      - name: Set up Python 3.12
         uses: actions/setup-python@v3
         with:
-          python-version: '3.11'
+          python-version: '3.12'
       - uses: actions/setup-node@v3
```
Stormheg, line 22: This might be an excellent opportunity to upgrade the versions of the actions we use ;)

Stormheg (approved):
Gave this a go locally. I don't have a representative database backup so I cannot test every aspect, assuming someone will test this on staging (staging site appears to be in maintenance mode?)

It looks like none of the upgrade considerations in the 6.2 release notes affect us, so should be straightforward to upgrade.

We should be sure the production database is also on Postgres 16 or better before we release, [as noted by Dan](https://github.com/wagtail/wagtail.org/pull/475#issuecomment-2242667848)

vossisboss:
@Stormheg, our production database was force upgraded about a week or so ago, so we're definitely on the correct version.

I will give this one more look later today or tomorrow then merge this.

vossisboss:
@zerolab The bookworm build failed on staging. Here is the key error. You can poke at the full log on Heroku if you want. Have we used bookworm elsewhere with no issues? Just wondering if it's worth pushing this or if we should stick to bullseye for this iteration of the site.

```
E: Unable to locate package postgresql-client-13
The command '/bin/sh -c apt-get update --yes --quiet &&     apt-get install -y apt-transport-https rsync libmagickwand-dev unzip postgresql-client-13     jpegoptim pngquant gifsicle libjpeg-progs webp &&     rm -rf /var/lib/apt/lists/*' returned a non-zero code: 100
```

cnk:
I find [this page a little confusing ](https://packages.debian.org/search?keywords=postgresql-client)but I suspect what we want is `postgres-client-15` to go with the newer base image

vossisboss:
I'm not sure where and how we would set that @cnk? The Dockerfile? Looks like the bullseye deployment is working fine, so I'm not sure I'm really inclined to dig into this unless there are some amazing reasons to use bookworm.

Stormheg:
@vossisboss is pushed a commit that should fix the dockerfile.

'bullseye' and 'bookworm' are - Toy Story inspired - codenames for Debian, the operating system the Docker containers use. Bullseye (Debian 11) is not receiving security updates anymore: https://endoflife.date/debian

cnk:
@vossisboss I tried @Stormheg's update from an hour ago and it works great. I had to export and reimport my database to upgrade the postgres server version but that was the only extra step to run this upgrade vs current main branch. I think this is ready to go to the Heroku staging server when you have a chance.

vossisboss:
Yeah, a lack of security updates is DEFINITELY a good reason to switch then. Testing this on staging. Then hopefully we can get this up on production. BTW, I think @cnk has staging access. @Stormheg if you want staging access, I can get that set up for you.

PR #479, opened 2024-11-01
Node 22 / Sass 1.80

Stormheg:
This upgrades the frontend toolchain to use the new Node 22 LTS and more modern versions of Sass and Stylelint.

There are a couple deprecations and new Stylelint rules that require some effort to remediate. I've decided to ignore those changes for now.


thibaudcolas (approved):
All looking great!

PR #478, opened 2024-10-30
Upgrade to Wagtail 6.3

Stormheg:
- Wagtail 6.3.0
- Django 5.0.9
- Willow 1.9.0
- Pillow 11.0
- Whitenoise 6.8 (nice performance improvements in this release to `collectstatic`!)
- Removal of the unused `alternative_text` field, use the new `description` field from Wagtail 6.3 instead.

PR #480, opened 2024-11-14
Migrate some `ImageChooserBlock` instances to `ImageBlock`

Stormheg:
Commit extracted from https://github.com/wagtail/wagtail.org/pull/478 (minus migration files to avoid conflicts)

We are not ready to move our usage over yet, as this will immediately result in validation errors when an editor tries to save an existing page with instances of `ImageBlock` on them, as the contextual alternative text will be missing

This PR is pending a decision on the way forward.

cc @cnk @vossisboss


cnk:
Hmmm perhaps I should remove the commit with the patch to use default alt text. It works great in the browser and most titles I checked were perfectly fine as the image's alt text.

I tried to look into the failing python tests but I can't run them locally. I have a virtual env and tried `poetry run ./manage.py test` (copied from ci.yml) but I get the same error as in CI for all branches I tried - including `main`

cnk:
I rebased this onto current main and whatever was causing tests to fail appears to have resolved itself. 

However, I am not getting the behavior I expect when creating a new blocks which contain an ImageBlock. If I choose an image I know has a description, I expected to see the "Alt text" field filled in with that description - like the image chooser in Draftail has done all along. But I am just getting blank fields for 'decorative' and 'alt/description'. If someone can replicate this on their installs, I think I'll raise an 'am I mistaken' issue on wagtail itself. @Stormheg @vossisboss 

(Note: I added description text on my local copy of the site; YMMV with other copies of the DB)
Screenshot 2024-12-04 at 3 27 33 PM

Chiemezuo:
@cnk @Stormheg 
Since we've updated the Wagtail version, I'm going to give this a go if that's okay by you both.

vossisboss:
No need to ask for permission @Chiemezuo. If you got time to move something forward, go ahead and do it.

Stormheg:
Closing this in favor of #503

PR #488, opened 2024-12-17
Switch to ruff for linting

zerolab:
Builds on #486 and switched the Python linting to Ruff


ruff.toml
```diff
@@ -0,0 +1,35 @@
...
+    "C4",   # flake8-comprehensions
+    "DJ",   # flake8-django
+    "E",    # pycodestyle errors
+    "F",    # pyflakes
+    "I",    # isort
+    "INT",  # flake8-gettext
+    "PIE",  # flake8-pie
+    "PGH",  # pygrep-hooks
+    "S",    # flake8-bandit
+    "SIM",  # flake8-simplify
```
Stormheg, line 16: Most of these I had never heard of but solve things that semi-regularly come up in code review I do.

I certainly will be adopting this ruff config ❤️

wagtailio/areweheadlessyet/migrations/0001_initial.py
```diff
@@ -1,11 +1,11 @@
 # Generated by Django 3.2.8 on 2022-01-24 11:57
 
 import django.db.models.deletion
-from django.db import migrations, models
-
 import wagtail.blocks
 import wagtail.fields
 
+from django.db import migrations, models
+
```
Stormheg, line 9: @zerolab this change looks a bit strange to me - I assume this happens because `import x` takes precedence over `from x import y`. I would prefer third-party imports stay in alphabetical order.

Is this intentional behavior by ruff or is this something we can have an opinion about?

Stormheg (commented):
@zerolab looks good but have one question/suggestion.

wagtailio/areweheadlessyet/migrations/0001_initial.py
zerolab: it is a ruff opinion. will have a look tommorrow whether we can influence that

zerolab: the best I could get it to is:

```diff
+ import django.db.models.deletion
from django.db import migrations, models
- import django.db.models.deletion

import wagtail.blocks
import wagtail.fields
```

using https://docs.astral.sh/ruff/settings/#lint_isort_force-sort-within-sections

Stormheg (approved):
That's acceptable to me 👍

PR #496, opened 2025-01-29
Dumpdata - does not work (for Meagen)

Stormheg:
Missing revisions cause the dump to not restore!


Stormheg:
Cleaning up revisions before dumping using `manage.py purge_revisions` and including it in the dump helps, but then we get stuck on foreign key references to users that are not part of the dump...

PR #498, opened 2025-02-07
Cleaning out unecessary tables

vossisboss:
In the course of cleaning up Wagtail.org, @cnk and I discovered that some unneeded tables were left behind in the database when the AB Testing and Google import packages were removed. As long as these tables are present, we can't delete any old images or data that are connected to them.

This PR adds a migration that removes three unnecessary tables from the database. Once these tables are removed, we can delete the extraneous data from the website.


cnk (changes requested):
The mismatch between number of forward statements and reverse statements bothered me - but doesn't bother Django. Since we don't have to have the same number of items in each list, I suggest we change the reverse_sql line to simply `reverse_sql=[migrations.RunSQL.noop]`

This works on my local copy of wagtail.org - in either configuration.

wagtailio/core/migrations/0061_remove_excess_tables.py
```diff
@@ -0,0 +1,25 @@
...
+
+    operations = [
+        migrations.RunSQL(
+            elidable=True,  # This migration can be removed during a squash
+            sql=[
+                "DROP TABLE IF EXISTS wagtail_ab_testing_abtesthourlylog;",
+                "DROP TABLE IF EXISTS wagtail_ab_testing_abtest;",
+                "DROP TABLE IF EXISTS wagtail_image_import_driveidmapping;",
+            ],
+            reverse_sql=[migrations.RunSQL.noop, migrations.RunSQL.noop],
```
Stormheg, line 23: Per @cnk's comment

```suggestion
            reverse_sql=[migrations.RunSQL.noop],
```

wagtailio/core/migrations/0061_remove_excess_tables.py
```diff
@@ -0,0 +1,25 @@
...
+    The Wagtail AB Testing module as well as the Google import module were removed from this project without dropping its tables.
+    """
+
+    dependencies = [
+        ("core", "0060_alter_contentpage_body_alter_homepage_body"),
+    ]
+
+    operations = [
+        migrations.RunSQL(
+            elidable=True,  # This migration can be removed during a squash
```
Stormheg, line 17: Yay! Suggesting this flag is a comment I often make during reviews of manual migrations like this one - but you are already way ahead here 🥇

vossisboss:
Well everyone seems happy, so I'm going to go ahead and merge this! Thanks y'all!

PR #503, opened 2025-02-26
Switch to ImageBlock

Chiemezuo:
This is a follow-up to #480 

I removed the patch (and the resulting older migrations) as the issues it addressed were fixed in Wagtail 6.4.

I also rebased so I could preserve the commit authors and their history, with a few additions.


vossisboss:
After testing all the blocks where `ImageBlock` occurs locally, it seems every block that includes the new `ImageBlock` is functioning properly. There were no issues with saving, updating, or publishing any images with the new version of the block. The only issue, which is an ongoing bug that needs to be solved in core, was the repeating of the word "Image" in blocks that were built using `StructBlock`. Like in this screenshot here:

Screenshot 2025-03-05 at 4 42 10 PM

The benefits of implementing the new block outweigh the issue of having the word image appear twice in my opinion. I think implementing this block to improve our alt text is more important. This issue needs to be properly logged and addressed in future versions of Wagtail Core though.

vossisboss:
I've asked @Chiemezuo to rename the migration files so that they better describe the change being made with each migration.

wagtailio/core/blocks.py
vossisboss, line 1: Looks like there is an instance of ImageChooserBlock left in this file that wasn't replaced by the new ImageBlock. Did you not change it because it wasn't changed in the prior PR @Chiemezuo? Or do you have another reason for leaving it be?

See line 109 in the LogoCardBlock.

wagtailio/services/blocks.py
vossisboss, line 1: Looks like there is another instance of ImageChooserBlock on Line 22 under CardBlock. Is there a reason we're not changing this one? If not, we should probably see if the new ImageBlock works well here.

vossisboss (changes requested):
All right @Chiemezuo, after doing my initial testing and review, there are a few things that need to be addressed.

As we discussed on the call, it would be really helpful if all the migration files had a more descriptive change to let future maintainers know what changes were made by the migration.

I also found a few instances where ImageChooserBlock was not updated to the new ImageBlock and I would like to know whether those were deliberate choices or whether there were a few instances you missed. If you could go through those and also search the project files for remaining instances that need to be replaced, that would be great.

I am also curious about the update needed to our custom image model. I'd just like to know a bit more about why it's necessary. And we'll want to test this on staging.wagtail.org for sure to make sure any changes to the model don't lead to database problems with our existing images.

Chiemezuo:
> I am also curious about the update needed to our custom image model. I'd just like to know a bit more about why it's necessary. And we'll want to test this on staging.wagtail.org for sure to make sure any changes to the model don't lead to database problems with our existing images.

For this, it's a minor change. The AbstractImage model was modified around the same time that the ImageBlock was being created. It adds a 'description' field which will be the primary default for images that do not have alt text.

Prior to the new changes, the alt text default was the 'title' field. However, with the new changes, it will be alt text -> description -> title, in that order of defaults.

vossisboss (approved):
All right! Testing on local and staging went very well. I discovered one unrelated bug but otherwise I think this will be a good update to our website.

vossisboss:
One note for us to remember after merging this: The new description fields Chieumezo added will need to be updated for all snippet images since most of those models don't incorporate `ImageBlock`. We'll have to make sure the description field fall backs work well for our alt text overall.

PR #504, opened 2025-02-26
Added an image usage report

cnk:
This is a report for identifying images that are not currently in use. Right now it is very basic - a list of all images on the site and a count of how often they are used. There are links to the image usage page and to the image edit where one can delete the image. It is usuable in its current condition but if anyone wants to extend this report, it might be useful to add a FilterSet that allows filtering by collection and/or by used / not used.

### DIFF
--- a/wagtailio/images/views.py
+++ b/wagtailio/images/views.py
@@ -1,5 +1,5 @@
 from django.contrib.contenttypes.models import ContentType
-from django.db.models import CharField, Count, IntegerField, OuterRef, Subquery
+from django.db.models import CharField, Count, OuterRef, Subquery
 from django.db.models.functions import Cast, Coalesce
 from django.utils.translation import gettext_lazy as _
 
@@ -39,8 +39,11 @@ class ImageUsageReport(ReportView):
                 # to_object_id is a varchar so we need to cast the image PK to a varchar for comparison
                 to_object_id=Cast(OuterRef("pk"), output_field=CharField()),
             )
-            .values(cast_to_object_id=Cast("to_object_id", output_field=IntegerField()))
-            .values(count=Count("id"))
+            .values("object_id", "to_object_id")
+            .distinct()
+            .values("to_object_id")
+            .annotate(count=Count("object_id", distinct=True))
+            .values("count")
         )
 
         return WagtailIOImage.objects.annotate(
--- a/wagtailio/images/views.py
+++ b/wagtailio/images/views.py
@@ -3,11 +3,58 @@ from django.db.models import CharField, Count, OuterRef, Subquery
 from django.db.models.functions import Cast, Coalesce
 from django.utils.translation import gettext_lazy as _
 
+from wagtail.admin.filters import CollectionFilter, WagtailFilterSet
 from wagtail.admin.ui.tables import TitleColumn
 from wagtail.admin.views.reports import ReportView
-from wagtail.models import ReferenceIndex
+from wagtail.models import Collection, ReferenceIndex
+
+from django_filters import ChoiceFilter as DjangoChoiceFilter
 
 from .models import WagtailIOImage
+
+
+class UsageFilter(DjangoChoiceFilter):
+    def filter(self, qs, value):
+        # Value passed will be one of the choices
+        if value:
+            if value in ["0", "1"]:
+                return qs.filter(usage_count=value)
+            else:
+                # If the value is not 0 or 1, we can assume means greater than 1
+                return qs.filter(usage_count__gt=1)
+        # If the value is not in the choices, return the original queryset
+        return qs
+
+
+class ImageUsageReportFilterSet(WagtailFilterSet):
+    collection = CollectionFilter(
+        label=_("Collection"),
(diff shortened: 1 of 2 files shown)


cnk:
I fixed the sorting issue but am still having trouble changing the query to count how many things reference an image rather than the total number of references. A page using an image as its SEO media image and in the body of the page is counted as 2 references in this query while the number displayed on the image edit page would only count that as 1 use.

cnk:
More info for the "help wanted". The query I am using for this report counts all the lines in the Reference Index that reference a given image. But the usage display on the image edit page shows the count of items referencing the image. These two counts differ when a page uses an image multiple times - for example in the text and as the social media image. This is essential a Django/Wagtail puzzle. You don't need to have the data from wagtail.org to be able to contribute to this issue because you can copy the report into any Wagtail project you have set up and figure out the query there.

cnk:
Discussion for moving the information in this report into the image listing views: https://github.com/wagtail/wagtail/issues/12550

wagtailio/images/views.py
```diff
@@ -0,0 +1,48 @@
...
+    ]
+
+    def get_base_queryset(self):
+        # Create the subquery to count matching ReferenceIndex rows
+        reference_index_subquery = (
+            ReferenceIndex.objects.filter(
+                to_content_type=ContentType.objects.get_for_model(WagtailIOImage),
+                # to_object_id is a varchar so we need to cast the image PK to a varchar for comparison
+                to_object_id=Cast(OuterRef("pk"), output_field=CharField()),
+            )
```
joelwilliam2005, line 41: You can add:
```
.values('content_type_id', 'object_id')
.distinct()
```

This approach will create a pair of images and where the image is used. 
And even if an image is used two times in a page, you will have one referencing item as using distinct() will ensure each item is counted only once.

cnk: @joelwilliam2005 thanks for the attempt. I am not 100% sure where you want me to add those lines, but every place I tried, I get `ProgrammingError at /admin/reports/image_usage_report/
more than one row returned by a subquery used as an expression`

The version that doesn't error (below) this gives the exact same counts as the version in the current PR. 

```reference_index_subquery = (
            ReferenceIndex.objects.filter(
                to_content_type=ContentType.objects.get_for_model(WagtailIOImage),
                # to_object_id is a varchar so we need to cast the image PK to a varchar for comparison
                to_object_id=Cast(OuterRef("pk"), output_field=CharField()),
            )
            .values('to_content_type_id',  cast_to_object_id=Cast("to_object_id", output_field=IntegerField()))
            .distinct()
            .values(count=Count("id"))
        )
```

cnk: Thanks. That doesn't error - and doesn't change the query results. It is still counting all the lines in the ReferenceIndex table.

joelwilliam2005: I'll look into it.

joelwilliam2005: Can you try this approach and see if it works?
```
def get_base_queryset(self):
        # Create the subquery to count distinct items referencing each image
        reference_index_subquery = (
            ReferenceIndex.objects.filter(
                to_content_type=ContentType.objects.get_for_model(Image),
                to_object_id=Cast(OuterRef("pk"), output_field=CharField()),
            )
            .values("object_id", "to_object_id")  
            .distinct()  
            .values('to_object_id')  
            .annotate(count=Count('object_id', distinct=True))  
            .values('count') 
        )
```

joelwilliam2005: You'll have to import this as well:
`from wagtail.images.models import Image`

cnk: YES!!!!!!! @joelwilliam2005 can you please push a second commit to this PR branch - but keep WagtailIOImage rather than Image in the to_content_type line? I want you to add this as a commit so it's in your GH history rather than mine.

joelwilliam2005: Awesome!! Glad to hear that it works.
I've created a PR #509 for it.

cnk:
@vossisboss I think this report is ready to go to production. The counts now match the usage data you get for an individual image and I have added a filter so you can find images used, 0, 1, or multiple times. You can then export them. The export includes the title and description in case you want to use it to prioritize better alt text for images in use.

Screenshot 2025-05-21 at 4 48 34 PM

vossisboss:
This looks good! Thank you so much for contributing this report @cnk! I think this one would be a great one to include in our standard reports too if you want to propose adding it to Wagtail core as well.

laymonage (commented):
I tried this at https://wagtail.org/admin/reports/image_usage_report/ but it throws a 500 error

```
ModelChoiceField.__init__() missing 1 required positional argument: 'queryset'
```

Sentry issue (if you have access):
https://torchbox.sentry.io/issues/6640591798/?environment=production&project=1220804&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream&stream_index=0

I wonder if this broke after #521 or not. I'll try to look into it when I have time.

PR #506, opened 2025-03-06
Remove taggit items from our ReferenceIndex

cnk:
We have been cleaning up unused images and are seeing some odd items in our image usage reports. When I investigated, the "private tagged item" comes from the taggit tags. I think these got added when we upgraded to Wagtail 4.1. [As of Wagtail 5.0](https://docs.wagtail.org/en/stable/releases/5.0.html#referenceindex-no-longer-tracks-models-used-outside-of-wagtail), the reference index no longer tracks non-Wagtail model references. We can safely remove them. That's what this PR does.

Before running this migration, this image has 2 tags and says it is used 3 times while really it is only used once.

Screenshot 2025-03-05 at 6 08 44 PM

After running this migration, we still have the 2 tags but we see the correct "Used 1 time".

Screenshot 2025-03-05 at 6 25 42 PM


laymonage (commented):
Couldn't this be done by running the `rebuild_references_index` management command? The reference index [deletes existing references](https://github.com/wagtail/wagtail/blob/dbf920e00c77844152c51569de49d5543164edb0/wagtail/management/commands/rebuild_references_index.py#L39-L43) before rebuilding it, so it should remove the stale references as well.

Looking at our Heroku dashboard, looks like we haven't added it to our scheduler. 
image

I'll add it and run it manually once, and see if it fixes the issue.

laymonage (commented):
Looks like that did the trick!

image

image

I'll close this PR then. Thanks for working on this and bringing it to our attention!

PR #511, opened 2025-03-25
Fix #508: Ensure consistent animations for logo cards

### LINKED ISSUE #508: Animation bug on Services page
When adding a new card within the Logos block on the Services page, the new block doesn't have the same animation styling as the old block. Not sure why that's happening, so we should probably investigate and/or revamp the Services page.

Image

Rish-it:
- Add position:relative to base logo-card class

- Add transform transition for smoother animations

- Move animation styles out of :is(a) selector

- Apply animations consistently to all logo cards

### DIFF
--- a/wagtailio/static/sass/components/_logo-card.scss
+++ b/wagtailio/static/sass/components/_logo-card.scss
@@ -13,6 +13,8 @@
     background-color: var(--color--background);
     padding: 20px;
     left: -20px;
+    position: relative;
+    transition: transform 0.3s ease;
 
     &__meta {
         margin-bottom: 10px;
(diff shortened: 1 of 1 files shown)


vossisboss:
Hello @Rish-it. I tested your proposed fix in both Firefox and Chrome. Unfortunately, it does not seem like the animation issue has been resolved. Here are some screenshots.

Screenshot 2025-03-26 at 6 15 10 PM

Screenshot 2025-03-26 at 6 13 05 PM


It's probably hard to test the fix yourself without access to the backend or test data for the Services page, so if you want to join our Slack community and coordinate with us in the #wagtail-org channel, we can probably get you paired up with someone who can help you with testing.

Please do give this another go if you want to and resubmit if you have another idea for how to fix this bug.

Rish-it:
@vossisboss  Thanks for testing the fix and sharing the screenshots. I appreciate the feedback. I’ll look and see if I can find a better solution. Since I’m already in the Slack channel, could you pair me up with someone who can help with testing? That would make debugging a lot easier.

vossisboss:
Hi @Rish-it. I actually discovered that something was up with my `npm`server. So it looks like this fix is actually functional. My bad! Let me take a deeper look at this fix and see whether the changes you're proposing have any unexpected effects on other components.

I'll still connect you with some other folks in the Slack channel just in case you spot any other issues you want to try and fix.

Rish-it:
@vossisboss Hey! No problem at all. Just checking in—did you get a chance to review the fix? Let me know if you ran into anything unexpected. Also, check your Slack DM when you get a chance!

PR #512, opened 2025-03-26
make an experimental change

Chiemezuo:
This tests an experimental swap from `image` tags to `picture` tags introduced in v5.2

This change will happen for main images in blogs.


Chiemezuo:
@vossisboss

Issue #348, opened 2022-11-18
Blog posts need to be able to have multiple authors


vossisboss:
We need to be able to accommodate blog posts that have more than one author. Ideally, posts would be able to accommodate up to three co-authors. Adding additional authors will impact our designs on both `Blogpage` and `BlogPageIndex`, so we need to consider the best approach both from a design and a model perspective.

fyunusa:
@vossisboss should I give this a try ?

fyunusa:
ok @vossisboss thanks for the heads-up

PR #513, opened 2025-03-27
Add Support for Multiple Authors in Blog Posts

oz456:
In Progress #348

### DIFF
--- a/wagtailio/blog/models.py
+++ b/wagtailio/blog/models.py
@@ -36,7 +36,8 @@ class BlogIndexPage(Page, SocialMediaMixin, CrossPageMixin):
         return (
             BlogPage.objects.live()
             .descendant_of(self)
-            .select_related("author", "author__image", "category")
+            .select_related("category")
+            .prefetch_related("blog_page_authors__author")
             .order_by("-date", "pk")
         )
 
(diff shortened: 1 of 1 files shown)


zerolab:
Note: this will need a migration from the current setup to the new one, so that no author data is lost

oz456:
Thanks @zerolab!! Updated it as needed.

oz456:
Stacking authors was making the layout a bit complex, so I arranged them in a column on smaller viewports. @vossisboss, as you mentioned, having three authors is likely a rare case, so this shouldn’t cause any major issues.
Let me know if any changes are necessary.

**Note**: This migration cannot be directly reverted back once applied.

vossisboss:
The design and the frontend look pretty good @oz456. Like Dan said, we needed a migration operation to preserve the author data because the way you wrote this PR doesn't move the author data we already have into the first author position. I have re-based your PR and added an operation in the migration file to move the data into the new model before it is removed. I'm going to test this in staging to see if that migration works and then we'll go from there.

vossisboss:
All right! Looks like everything worked well on staging. Since I am not the most experienced person with migrations, I think it would be wise to get one more review before we push this to production. @Stormheg or @cnk, either of you have time to give this a look?

Issue #463, opened 2024-02-26
Increase logo visibility in contrast themes / high-contrast mode


thibaudcolas:
Follow-up to #454. The logos of Wagtail users visible on the site homepage don’t display too well in Windows high-contrast mode / contrast themes. This is particularly problematic for the Royal College of Arts logo, which is fully invisible when displayed in a high-contrast "dark" theme.

I think the ideal fix would be to apply similar styles to the ones used for the logos in the site’s dark theme:

https://github.com/wagtail/wagtail.org/blob/e8d45141c6f1e6bc97b4dead85cb560cfab9710a/wagtailio/static/sass/components/_logos.scss#L61-L68

By adding a `forced-color-adjust: none` alongside the background color, we’d ensure the logos are always displayed with the intended background color. It seems to me like all those styles could be present for all themes rather than forced-colors mode / the dark theme only, so the logos display as consistently as possible.

kituuu:
Can I fix this? But I don't have a windows system to check this.

GajulaSamatha:
Hii sir,
I want to work on this issue, and I need some clarification. Please feel free to clarify.

1) 
![Image](https://github.com/user-attachments/assets/eb03b3a7-a162-4566-942d-faa61434ba3b)


Are you referring to this logo? 

2) Or is it the following one?

![Image](https://github.com/user-attachments/assets/d3459264-4ea2-4f6b-903f-508826a738e5)

Thank you in advance sir.

GajulaSamatha:
Hello sir,

I have now cleared my doubts (as you suggested, the changes should be in logos.scss, not the Wagtail header logo) and attempted to solve the issue. To confirm, I am attaching screenshots of the output:

1) White-Contrast in system with Light Mode in website.

![Image](https://github.com/user-attachments/assets/2b8b7cc0-fa55-416e-b6c4-f7e4363054d6)

2) White-Contrast in system with Dark Mode in website.

![Image](https://github.com/user-attachments/assets/800778f2-681b-4faf-8d0b-4d0824be830d)

3) Dark-Contrast in system with Dark Mode in website.

![Image](https://github.com/user-attachments/assets/e71b3738-56a5-48ea-96c1-36d64664c100)

4) Dark-Contrast in system with Light Mode in website.

![Image](https://github.com/user-attachments/assets/f91d3a99-cbd5-41ee-ad9e-c24302d8b695)

I tested this in Chrome, Microsoft Edge, and Firefox.

[...]

PR #516, opened 2025-04-09
Fix: Increase logo visibility in contrast themes / high-contrast mode #463

GajulaSamatha:
This is my first PR, so if I made any mistakes, please excuse me and kindly guide me in the right direction. I hope to learn something new through this! 

The issue is #463 

To solve this issue, I followed @thibaudcolas sir's suggestion of adding" forced-color-adjust: none". It successfully worked when switching to system's high-contrast mode from website's dark mode.

However, while testing in light mode, there was no specific background color applied to the logo, which made it easy for contrast themes to override it. So, I explicitly added background-color: white here (these styles are applied in light mode):

image

these are the styles applying in the light mode of the website.

During testing, I also observed that PNG format logos are better in such cases, since their background color can be styled easily. For other formats, the appearance may not be ideal. You can see the difference here:

image

The first logo is a PNG.
The second one is not, and does not blend well.

I tested this in Chrome, Microsoft Edge and Fire Fox browsers.

### DIFF
--- a/wagtailio/static/sass/components/_logos.scss
+++ b/wagtailio/static/sass/components/_logos.scss
@@ -49,6 +49,8 @@
         min-width: 100px;
         max-width: 100px;
         flex-basis: 100px;
+        background-color: $color--white;
+        forced-color-adjust: none;
 
         .theme-dark & {
             min-width: 120px;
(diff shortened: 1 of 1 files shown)


thibaudcolas:
spot on, ty!

PR #518, opened 2025-04-14
Refactor social sharing meta

### LINKED ISSUE #514: Clean up image sharing sizes to render on LinkedIn
Our [social media sharing image metadata](https://github.com/wagtail/wagtail.org/blob/cd0def555e35b36aae8b440640d11e33ee8a68c4/wagtailio/project_styleguide/templates/patterns/base_page.html#L14-L43) seems to not be set up properly.

Example post: [OLED and dark websites = lower footprint ✨](https://wagtail.org/blog/oled-and-dark-websites-lower-footprint/). The [source image](https://media.wagtail.org/images/crop_wagtail-dark.original.jpg) is 1600x900.

- The Twitter image is 400px. I’m not clear why we would be using a different social image just for that network. If we keep the meta tag, we probably should use the same image rendition as `og:image`.
- The OpenGraph image is 1114 x 627, as it’s rendered with `width-1200 height-627`. This is below the [LinkedIn recommended minimum of 1200x627](https://www.linkedin.com/help/linkedin/answer/a521928)

I believe the problem here is there shouldn’t be two `width-` and `height-` tags in use. Images with an aspect ratio squarer than the recommended 1.91:1 will be resized below the minimum size.

- width-1200 takes the 1600x900 image to 1200x675
- height-627 takes the 1200x675 to 1114x627

Instead we probably want the `min` or `fill` filters.

oz456:
Fixes #514
Updated social sharing image dimensions using `fill-1200x630` to match recommended sizes across most of the platforms.

Additional Note:
Some blog posts currently do not meet the required minimum description length of 100 characters. we could consider adding `min=100` to the introduction field.
Not sure about all other platforms, but linkedin recommends a minimum of 100 characters.

### DIFF
--- a/wagtailio/project_styleguide/templates/patterns/base_page.html
+++ b/wagtailio/project_styleguide/templates/patterns/base_page.html
@@ -18,6 +18,23 @@
         {% image page.main_image fill-1200x630 as social_img %}
     {% endif %}
 
+    <meta property="og:title" content="{% if page %}{% firstof page.seo_title page.title %} | {% endif %}Wagtail CMS" />
+    <meta property="og:type" content="website" />
+    {% if page %}
+        <meta property="og:url" content="{% fullpageurl page %}" />
+    {% endif %}
+    {% if social_img %}
+        <meta property="og:image" content="{{ social_img.url }}" />
+        <meta property="og:image:width" content="{{ social_img.width }}" />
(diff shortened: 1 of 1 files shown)


thibaudcolas:
lgtm, ty!

PR #492, opened 2025-01-11
Feature/481 wagtail newsletter

mgax:
Fixes https://github.com/wagtail/wagtail.org/issues/481.

- [x] Install `wagtail-newsletter` and render some minimal MJML
- [x] Download newsletter archive from Mailchimp
- [ ] Replicate the recent "This Week in Wagtail" issues listed in https://github.com/wagtail/wagtail.org/issues/481#issuecomment-2484414303
- [ ] Validate rendered newsletters using Litmus
- [ ] Make sure the public view of a `NewsletterPage` looks good
- [ ] Fix migrations and preserve the existing newsletter page content

---
`wagtail-newsletter` recommends installing the `mrml` library to render MJML, but the `mrml` package doesn't work on Python 3.13 (https://github.com/wagtail/wagtail-newsletter/issues/69), so we use `mjml-python` (another wrapper of the same underlying Rust `mrml` library) for now.

Issue #468, opened 2024-03-18
Add a CSP to wagtail.org


thibaudcolas:
This would help protect the site against possible XSS issues and other types of injection (for example #169).

thibaudcolas:
There is a bit more needed (configuration of the server, removal of django-csp package), but #550 is definitely the next step.

PR #519, opened 2025-04-16
Implement wagtail-newsletter

### LINKED ISSUE #481: Port "This Week in Wagtail" to wagtail-newsletter
The newsletter used to be content-managed in wagtail.org until July 2023 (the latest page is https://wagtail.org/this-week-in-wagtail/issue-136/). Content display issues on Outlook prompted a move to a content template in Mailchimp.

Since https://github.com/wagtail/wagtail-newsletter is released, and it [comes with MJML support](https://wagtail-newsletter.readthedocs.io/en/stable/customization.html#newsletter-templates), we can reimplement the current Mailchimp template as MJML, and move back to Wagtail-managed newsletters.

mgax:
Fixes https://github.com/wagtail/wagtail.org/issues/481. This PR supersedes https://github.com/wagtail/wagtail.org/pull/492.

I was able to replicate a recent newsletter edition with very small spacing differences. I'm pretty happy with the editing experience and have somewhat tested this using Litmus. The main issue with testing is that images are served locally, so Litmus can't load them; we'd need to either deploy this to staging, or use a tunnelling tool like ngrok.

Things that I'm not so happy about:
- Links in the accent block aren't styled as white text in some Outlook versions.
- The vertical spacing is a bit fiddly, since I mostly had to work with padding, not margins (margins can overlap, padding can't).
- ~There's no code to migrate the archive from Mailchimp just yet. I'm hoping that an AI might make writing that code not too painful.~ Import works.
- ~The detail view of a newsletter will load the email-formatted HTML from a variable into the shadow DOM of a`<div>`, so that its styling doesn't break the header and footer. I'm not sure how well this plays with search engines.~ The content is now rendered using the [Declarative Shadow DOM](https://developer.mozilla.org/en-US/docs/Web/API/Web_components/Using_shadow_DOM#declaratively_with_html) syntax.
- ~The `import_newsletter_archive` command doesn't work for TWiW issues 138 through 172.~ Fixed.


mgax:
I've added a management command to import a newsletter edition:

```shell
./manage.py import_newsletter https://mailchi.mp/wagtail/twiw-11034221 "Issue #180"
```

It will skip over the date and the footer, detect headings, copy images, and preserve links in the text. It doesn't copy the "Packages" section using the right block type (`AccentRichTextBlock`) but I'm wondering if that might be best cleaned up manually.

vossisboss:
FYI Alex. I got all the packages installed and such to test your branch but I'm running into a migration error that reads:

`cannot ALTER TABLE "newsletter_newsletterpage" because it has pending trigger events`

Baptiste suggested that we split one of the migration files in two. Apparently Postgres doesn't like it when you try to combine certain actions together. I'm going to attempt that and see if it fixes the issue. If I have no luck, I may need you to take another look at this or help me through the fiddly migration parts.

mgax:
I couldn't replicate the migration error, but I've encountered such errors in the past, and I think it's best to avoid them. I've split the migration into two steps.

wagtailio/newsletter/migrations/0005_wagtail_newsletter_data.py
```diff
@@ -0,0 +1,42 @@
...
+        ),
+        migrations.RemoveField(
+            model_name="newsletterpage",
+            name="intro",
+        ),
+        migrations.RenameField(
+            model_name="newsletterpage",
+            old_name="content",
+            new_name="body",
+        ),
```
laymonage, line 41: I'd recommend splitting the `RemoveField` operations to another migration. The problem tends to happen if you have both a data migration operation (in this case `RunPython(convert_body_to_streamfield)`) and a schema migration operation in a single migration. Migrations are run in a single transaction (if the backend supports DDL transactions), and the database may not allow data manipulation in a DDL transaction.

mgax: Fair point. I've also noticed that images were not getting preserved in the data migration so I've fixed that too.

wagtailio/newsletter/models.py
```diff
@@ -3,39 +3,67 @@
...
+from wagtailio.newsletter.blocks import NewsletterContentBlock
 
-class NewsletterPage(Page):
-    date = models.DateField("Newsletter date")
-    intro = RichTextField(blank=True)
-    body = RichTextField()
+
+@register_setting
+class NewsletterSettings(BaseGenericSetting):
+    footer = StreamField(NewsletterContentBlock(), blank=True, use_json_field=True)
```
Stormheg, line 18: @mgax this and the StreamField below don't require setting `use_json_field` anymore, see https://docs.wagtail.org/en/stable/releases/6.0.html#streamfield-no-longer-requires-use-json-field-true


```suggestion
    footer = StreamField(NewsletterContentBlock(), blank=True)
```

wagtailio/newsletter/models.py
```diff
@@ -3,39 +3,67 @@
...
+
+    panels = [
+        FieldPanel("footer"),
+    ]
+
+
+class NewsletterPage(NewsletterPageMixin, Page):
+    date = models.DateField()
+    preview = models.TextField(blank=True)
+    body = StreamField(NewsletterContentBlock(), blank=True, use_json_field=True)
```
Stormheg, line 28:
```suggestion
    body = StreamField(NewsletterContentBlock(), blank=True)
```

wagtailio/newsletter/models.py
```diff
@@ -3,39 +3,67 @@
...
         FieldPanel("date"),
-        FieldPanel("intro"),
+        FieldPanel("preview"),
         FieldPanel("body"),
     ]
 
     search_fields = Page.search_fields + [
         index.SearchField("intro"),
         index.SearchField("body"),
+        index.SearchField("body"),
```
Stormheg, line 39: Accidental double?

```suggestion
```

wagtailio/newsletter/templates/newsletter/newsletter_mjml.html
```diff
@@ -0,0 +1,96 @@
...
+            </mj-section>
+
+            <mj-section padding="6px 52px 30px">
+                <mj-column>
+                    <mj-text css-class="rich-text">
+                        <p>
+                            Want to change how you receive these emails? You can
+                            <a href="*|UPDATE_PROFILE|*">update your preferences</a> or
+                            <a href="*|UNSUB|*">unsubscribe from this list</a>.
+                        </p>
```
Stormheg, line 84: **Suggestion:** it looks like these end up on the web page version. Shouldn't these links be hidden? They don't go anywhere.

image

wagtailio/newsletter/templates/newsletter/newsletter_mjml.html
```diff
@@ -0,0 +1,96 @@
...
+                <mj-preview>{{ page.preview }}</mj-preview>
+            {% endif %}
+        </mj-head>
+
+        <mj-body width="660px" background-color="#f4f4f4">
+            <mj-section padding="13px 0" background-color="">
+                <mj-column>
+                    <mj-text align="center" css-class="rich-text">
+                        <a href="{{ page.full_url }}">View this email online</a>
+                    </mj-text>
```
Stormheg, line 30: Same as below - shouldn't this be hidden on the web version of the page?

wagtailio/newsletter/templates/newsletter/newsletter_page.html
```diff
@@ -1,985 +1,17 @@
...
+    {{ email_html|json_script:"email-html-data" }}
+    <script>
+        (function() {
+            const container = document.getElementById("newsletter-container");
+            const shadow = container.attachShadow({mode: "open"});
+            const wrapper = document.createElement("div");
+            wrapper.innerHTML = JSON.parse(document.getElementById("email-html-data").textContent);
+            shadow.appendChild(wrapper);
+        })();
+    </script>
```
Stormheg, line 16: **Suggestion:** maybe there is an alternative for the shadow dom?

**Issue:** otherwise this should be converted to a dedicated JS script that is loaded on this page and creates the shadow DOM.

Adding inline JS is not acceptable as we're trying to add a Content Security Policy (see https://github.com/wagtail/wagtail.org/issues/468) in the future, which does not allow for inline JS

Stormheg (commented):
@mgax couple minors comments.

Impressed by the migration of the old issues (Issues <= 138). Though this is not how they originally looked, though I don't think it is super important to keep their original look. They look clean now.


> The detail view of a newsletter will load the email-formatted HTML from a variable into the shadow DOM of a<div>, so that its styling doesn't break the header and footer. I'm not sure how well this plays with search engines.

Have you considered loading the email HTML in an iframe? That should accomplish mostly the same thing. Iframes have been around for so long it is more likely they behave nicely in the context of SEO - if that is even important.

wagtailio/newsletter/templates/newsletter/newsletter_mjml.html
mgax: Fixed in bec1ad01.

wagtailio/newsletter/templates/newsletter/newsletter_page.html
mgax: Fixed in c65e242a. It turns out you can inline the shadow dom content.

vossisboss:
@mgax We may have skipped an issue. I wouldn't be surprised. It will just be a part of Wagtail lore now. I am testing your code right now. I pushed the Wagtail 7.0 upgrade today, so would you mind rebasing this to the latest version?

I don't have a lot more time left in my day to test this so I might pick this up again in the morning and send you my notes.

mgax:
> I pushed the Wagtail 7.0 upgrade today, so would you mind rebasing this to the latest version?

@vossisboss I did a merge with `main`, since it looks like we're squashing PRs anyway. I can rebase and force-push if that makes more sense.

mgax:
Hi @vossisboss, I have updated the PR:
- Merge with `main` and fix conflicts
- Fix a tiny bug – remove the `NewsletterPage.intro` field that was still listed in search fields
- Add recent TWiW issues to the list in `wagtailio/newsletter/management/commands/import_newsletter_archive.py`
- Have Claude update the HTML parser that converts old issues into clean StreamField blocks (some chunks of text and images were missing, and the Packages section was not rendered as an `AccentRichTextBlock`; now these are fixed)

I have spot-checked a few of the imported newer issues and they look the same as in the Mailchimp archive.

vossisboss:
Thanks for doing this @mgax. I figured we could use some of our time tomorrow to test this all on staging and see if there is anything we're missing.

PR #520, opened 2025-04-25
Django 5.2 upgrade

Stormheg:


Stormheg:
Turns out we are on 6.4, not the 6.3 LTS that just received official Django 5.2 support.

Going to close this in favor on an upgrade to Wagtail 7.0 as well, see #521

PR #521, opened 2025-04-25
Wagtail 7.0

Stormheg:
Haven't yet attempted to [mark certain model fields as nullable to test validation on save changes in 7.0](https://docs.wagtail.org/en/latest/releases/7.0.html#configuring-deferred-validation-of-required-fields) - assuming we can add those later as those changes are a bit more involved to implement.


vossisboss:
Hi @Stormheg. I poked around this branch during Office Hours and the only thing I have found so far is that the blog model needed a null setting on the date for deferred validation to work properly. Added a commit. I'll deploy this to staging and see if anything else pops up.

PR #524, opened 2025-05-07
Add block preview configuration for blocks used in blog app

laymonage:
Only the blog app for now, I'll let someone else take over the rest 🙂

Not sure if I can do anything about the white flash in dark mode, will look into it...

https://github.com/user-attachments/assets/0354fbd0-172f-4500-8cee-298f795cdcca


Stormheg:
Thank you for this Sage. First time I'm ever seeing the block preview feature in action 👍

PR #546, opened 2025-08-20
Wagtail Space Banner w/Conditional Template Logic

vossisboss:
Rather than using #545 to advertise Wagtail Space, I've created a banner using conditional template logic instead. The advantages of this approach are that we get to use images in the design and that the banner loads at the same time as the rest of the content on the page. The downsides are that it must be edited within the code base and we will have to remove this banner or comment it out in the code after Wagtail Space is over.

Any feedback or suggestions on this approach would be much appreciated!

### DIFF
--- a/wagtailio/project_styleguide/templates/patterns/base_page.html
+++ b/wagtailio/project_styleguide/templates/patterns/base_page.html
@@ -87,6 +87,18 @@
 {% endblock body_top %}
 
 {% block header %}
+    {% if '/wagtail-space' not in request.path %}
+    <!-- This content will NOT show on any URL containing /wagtail-space -->
+        {% if '/wagtail-space' not in request.path %}
+            <div class="space-banner">
+                <div class="banner-text">Join us for</div>
+                <img src="{% static 'img/WagtailSpace2025Logo.png' %}"
+                     alt="Registration banner"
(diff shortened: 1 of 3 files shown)


wagtailio/project_styleguide/templates/patterns/base_page.html
```diff
@@ -87,6 +87,18 @@
 {% endblock body_top %}
 
 {% block header %}
+    {% if '/wagtail-space' not in request.path %}
+    <!-- This content will NOT show on any URL containing /wagtail-space -->
+        {% if '/wagtail-space' not in request.path %}
```
zerolab, line 92: duplicated if? also `<!-- This content will NOT show on any URL containing /wagtail-space -->` should probably be `{# This content will NOT show on any URL containing /wagtail-space #}`

Also, is there a way to avoid hard-coded paths?

vossisboss: We could add some models to make the url paths controllable in the admin for sure. But since Thibaud and me are the primary users of this feature right now and we both have code base access, that might be overengineering for this year. This honestly may be what works the best for us for this year's event and we can add models into the mix for next year.

What are you doing commenting while on holiday? Miss us that much? 😛

PR #549, opened 2025-08-22
removed the unused-django-referer-policy package

### LINKED ISSUE #515: Migrate to Django SECURE_REFERRER_POLICY
The site currently uses the django-referrer-policy package, however this should no longer be needed.

Current implementation:

```python
# Referrer-policy header settings
# https://django-referrer-policy.readthedocs.io/en/1.0/

REFERRER_POLICY = env.get(
    "SECURE_REFERRER_POLICY", "no-referrer-when-downgrade"
).strip()
```

Once we migrate to the built-in support:

```python
# https://docs.djangoproject.com/en/stable/ref/middleware/#referrer-policy
SECURE_REFERRER_POLICY = env.get(
    "SECURE_REFERRER_POLICY", "no-referrer-when-downgrade"
).strip()
```

p-r-a-v-i-n:
This PR fixes #515 

According to the instruction mentions in given issue by author. I have removed the 'django-referrer-policy' dependency.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,7 +23,6 @@ django-csp = "~3.8"
 django-manifest-loader = "~1.0.0"
 django-pattern-library = "~1.5.0"
 django-permissions-policy = "~4.25"
-django-referrer-policy = "~1.0"
 django-storages = { version ="~1.14", extras = ["s3"] }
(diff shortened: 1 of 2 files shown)


Stormheg (approved):
LGTM

PR #550, opened 2025-08-27
Refactor to use upcoming Django 6.0 CSP support

### LINKED ISSUE #532: Refactor CSP configuration from django-csp to Django 6.0 core support
This is be a good test of the Django 6.0 capabilities, for @Chiemezuo to work on as part of the [Content Security Policy compatibility](https://github.com/wagtail/roadmap/issues/92) GSoC project.

Django 6.0 is only [due in December](https://code.djangoproject.com/wiki/Version6.0Roadmap), so this won’t be merge-able for a while, but this feels like a good task to do during GSoC nonetheless.

Chiemezuo:
Fixes #532

In December, [Django 6.0](https://docs.djangoproject.com/en/dev/releases/6.0/) is expected to be released.

One of the biggest features will be the built-in CSP support. This means that projects that previously used external CSP packages like [django-csp](https://django-csp.readthedocs.io/en/latest/) will no longer need to rely on external tooling. CSP will be supported out-of-the-box 🎊 

I've tested this tool, and I feel confident that we can replace our usage of django-csp with it. However, as Django 6.0 is not yet available, we cannot remove usage of django-csp.

This PR adds support for the coming Django 6.0 release, while keeping the existing django-csp functionality.

They will both be interchangeable, depending on whether it's a compatible Django version or not.


Chiemezuo:
Hi, @vossisboss 
I will add this to my todos and have some feedback possibly by weekend.

Issue #347, opened 2022-11-18
New block needed for image captions


vossisboss:
Our new design (and also our old design) didn't have very many good options for adding captions to images. We need either a new type of image block or a new type of text block to provide captions to images that need them. Whichever option we pursue will need to be accessible and provide a good experience for screenreader users.

Aarthy153:
Hi, I would like to work on this issue. Can I get assigned? I
f there is a screenshot mentioned for the exact location which needs to be fixed, it would be more helpful for me :)

PR #552, opened 2025-08-28
Add ImageCaptionBlock with optional caption support

p-r-a-v-i-n:
This adds a new ImageCaptionBlock that supports image upload with an optional rich-text caption (bold, italic, links).
It renders using `<figure>` and `<figcaption> `for better accessibility and screen reader support.

This PR addresses #347


vossisboss:
Hello @p-r-a-v-i-n. Thank you for this contribution. Due to the new way that Wagtail handles image descriptions and alt text in the default ImageBlock, we don't have a need for this feature anymore. I will close out the issue on this as well so that no one else picks it up.

If you want to connect with me via the Wagtail Slack, I'd be happy to help you find something else to try. Thanks again for sending this in!

PR #576, opened 2025-12-10
Updating cookie code and cookie message

vossisboss:
This PR contains updated cookie code that provides users with an option to opt in or opt out of our analytics tracking. This code will be more compliant with GDPR and will work with the new consent code that Google Analytics is providing.

We are going to be testing this code on staging first over the end-of-year holidays to make sure that these changes don't wreck the analytics we are currently relying on. This PR will be staying in the queue until then.

### DIFF
--- a/wagtailio/project_styleguide/templates/patterns/includes/cookie_message.html
+++ b/wagtailio/project_styleguide/templates/patterns/includes/cookie_message.html
@@ -1,14 +1,13 @@
 <div class="cookie" data-cookie-message>
     <div class="cookie__container">
         <div class="cookie__message">
-            <p>We use cookies to provide you with a better service. Carry on browsing if you're happy with this, or <a class="cookie__link" href="/cookie-policy/">find out more</a></p>
+            <p>We use cookies to measure how you use the website. We want our site to be easy for you to use; understanding how you interact with it helps us know that. <a aria-label="Find out more about how we use cookies" class="cookie__link" href="/cookie-policy/">Find out more</a>.</p>
         </div>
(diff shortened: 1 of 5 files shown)


vossisboss:
@thibaudcolas I've updated this code to use buttons instead. Please let me know if there's anything else. Neal is giving us his sysadmin side-eye, so if you could proritize this review, I'd appreciate it.

wagtailio/project_styleguide/templates/patterns/includes/cookie_message.html
```diff
@@ -1,14 +1,11 @@
 <div class="cookie" data-cookie-message>
     <div class="cookie__container">
         <div class="cookie__message">
-            <p>We use cookies to provide you with a better service. Carry on browsing if you're happy with this, or <a class="cookie__link" href="/cookie-policy/">find out more</a></p>
+            <p>We use cookies to measure how you use the website. We want our site to be easy for you to use; understanding how you interact with it helps us know that. <a aria-label="Find out more about how we use cookies" class="cookie__link" href="/cookie-policy/">Find out more</a>.</p>
```
thibaudcolas:
```suggestion
            <p>We use cookies to measure how you use the website. We want our site to be easy for you to use; understanding how you interact with it helps us know that. <a class="cookie__link" href="/cookie-policy/">Find out more</a>.</p>
```

it’s best we avoid overriding visible labels, so sighted users of assistive tech can trust what they’re looking at to match how their assistive tech understands the content of the site.

thibaudcolas (approved):
All looking good, working well!

PR #582, opened 2026-01-16
Fix mobile menu scrolling in landscape mode

### LINKED ISSUE #304: Mobile menu should be vertically scrollable for small screens
If I open the mobile menu on a device in landscape mode, the whole menu isn’t visible at once on the page, and can’t be scrolled. It should support scrolling.

AdityaSharma-Git3207:
Fixes an issue where the mobile navigation menu could not be fully scrolled in landscape mode, making the "Get started" button unreachable.

The mobile menu height is now constrained relative to the header height, enabling proper vertical scrolling when the menu is open.

Fixes #304

### DIFF
--- a/wagtailio/static/sass/layout/_header.scss
+++ b/wagtailio/static/sass/layout/_header.scss
@@ -85,7 +85,8 @@
             &.is-visible {
                 visibility: visible;
                 transform: translateZ(0);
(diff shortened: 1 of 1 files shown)


thibaudcolas:
thank you!

PR #584, opened 2026-01-20
Improved decorative shapes and headlines around it

AdityaSharma-Git3207:
This implements the remaining part of the suggested fix for #457.

At viewport widths ≤768px, section subheadings can be visually impacted by decorative shapes. This change adds a theme-aware, semi-transparent background behind the heading text on small screens to improve readability while preserving the decorative artwork.

I verified the change visually using browser DevTools in both light and dark modes.

Let me know if there are any changes that can be made.

### DIFF
--- a/wagtailio/static/sass/components/_headline.scss
+++ b/wagtailio/static/sass/components/_headline.scss
@@ -43,6 +43,19 @@
             padding-top: $gutter-lg;
         }
     }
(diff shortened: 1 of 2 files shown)


thibaudcolas:
you clearly haven’t tested this as there isn’t a `--color-background` in our code. So will close this now.

Issue #303, opened 2022-11-04
Logo should have all of its details preserved in dark forced-color themes


thibaudcolas:
Follow-up to #295. Our logotype is visible in high-contrast dark themes, but the details of the bird are lost. The logo’s full details should be visible across all contrast themes.

Scotchester:
Go for it, @satvikvash!

thibaudcolas:
Copying what @satvikvash and I discussed in DMs – here are the steps needed to get the site set up enough to work on this:

1. `npm install && npm run build`
2. createsuperuser
3. Log into the CMS
4. Delete the default homepage
5. Create a new Homepage
6. Create a new Site and assign it the homepage

dassudip2001:
it is still pending or fixes it ??

vossisboss:
@thibaudcolas Any recommendations on how to emulate this issues on non-Windows machines? @Stormheg is poking around BrowserStack, but it looks like we need a smidge of help recreating the issue. It looks like we've updated the logo since this issue was filed as well, so we need to do some fresh testing.

thibaudcolas:
I can confirm this is still an issue. There are two options if you want to try it out yourself:

1. In a Chromium browser, in the DevTools’ Rendering panel, set "Emulate CSS media feature forced-colors" to "forced-colors: active"
2.. Ask our accessibility team for access to our shared [Assistiv Labs](https://assistivlabs.com/) account, which supports trying out this theming as it’s implemented in Windows

vossisboss:
Thank you for the additional context and testing suggestions Thibaud!

PR #588, opened 2026-01-27
Improve Wagtail logo visibility in forced-colors mode

AdityaSharma-Git3207:
This fixes an issue where the Wagtail bird logo lost its internal details in forced-colors / high-contrast mode, causing it to appear as a flat shape. The change keeps the logo readable while leaving the rest of the UI unaffected (#303).


thibaudcolas:
this does not seem to do anything? will close as I expect you haven’t actually tested this?

PR #589, opened 2026-02-04
Fix mobile menu toggle on touch devices

AdityaSharma-Git3207:
Fixes #586 

Fixes the mobile menu toggle not closing on touch devices.

The toggle was bound to `mousedown`, which doesn’t reliably fire for subsequent taps on mobile browsers. This meant the menu could be opened but not closed again, leaving the navigation stuck open (and keeping page scroll locked).

This switches the event handler to `pointerup` for more reliable touch interaction.


thibaudcolas:
nice

PR #596, opened 2026-02-26
Update pre-commit hooks

Stormheg:


.pre-commit-config.yaml
```diff
@@ -3,18 +3,18 @@ default_language_version:
...
       - id: ruff-format
   - repo: https://github.com/pycontribs/mirrors-prettier
-    rev: v3.3.2
+    rev: v3.6.2
     hooks:
       - id: prettier
         types_or: [css, scss, javascript, ts, tsx, json, yaml]
   - repo: https://github.com/pre-commit/mirrors-eslint
-    rev: v8.20.0
+    rev: v10.0.2
```
thibaudcolas, line 17: this one requires more work

thibaudcolas (approved):
most of this is good to go but there are way more version numbers outside this file to keep in sync. And even in this file, for Node tooling the tag is meaningless if we don’t also update the dependency version resolution right below (fundamental flaw in how pre-commit works).

thibaudcolas:
Merged manually in 0e5e8dc0

PR #597, opened 2026-02-26
Use Debian Trixie as docker base

Stormheg:
Debian Bookworm will stop receiving security updates in June 2026.


vossisboss:
This seems to be working well on staging now. Apparently our choices were clean up our data or install a package to make Wagtail behave. This seems like the more reasonable approach for now.

PR #613, opened 2026-04-16
Archive of Wagtail Space 2025 subsite

cnk:
This creates a static version of the Wagtail Space 2025 subsite. 

There is a view that routes /wagtailspace/ to template files in the file system and I have persisted the images that would normally have been served from media.wagtail.org into this file system in case the images are changed in the main wagtail.org CMS. 

Comparing this version of the site to the actual site there are some differences that need to be reconciled - specifically the banner on the index page and the color of the register and schedule buttons in the middle of the page. 

I was also unsure how I wanted to correct the font includes in the CSS file so we'll need to figure out what would be the best option.

NOTE: I am using {% manifest %} for the JS files which means they will change when the main wagtail.org JS files change. Not sure if that is a bug or a feature. We should discuss and perhaps change that.


vossisboss:
I don't think including duplicates of the font files in the static folder is the end of the world since we'll be referencing them in future templates as well. I don't anticipate we'll be going nuts with a whole lot of new font files any time soon.

The colors of the buttons look okay to me? I think adding the rocketship banner back in was the main thing that needed to be fixed there.

What are the downsides of relying on `{% manifest %}`? The 2025 site didn't have any extra JS or animations but we might have some for 2026. Could we rely on a combination of `{% manifest %}` and adding extra JS as needed?

You also mentioned that needing to name the main template index.html was a limitation of this approach. I'm just wondering if there is a way to make the URLs pretty again. If there isn't, then this is something I think we can live with.

vossisboss:
P.S. We should definitely add a section to the README or create a separate MD file in the Wagtail Space static directory to document how you did this @cnk.

cnk:
> What are the downsides of relying on {% manifest %}?  

My concern about using the manifest links for the JS is that it would cause the JS in these archive files to change as the site evolves. It doesn't look like that would make any difference given the existing JS but I wanted to make a note of it in case someone could think of a reason not to do it. 

> I'm just wondering if there is a way to make the URLs pretty again.

The easiest way to pull down files with wget put  them all in directories + index.html files. I reprocessed the files to remove the directories, rename the files, and update the navigation links. If you would prefer to restore the original urls, e.g. https://wagtail.org/wagtail-space-2025/speakers/ instead of the new https://wagtail.org/wagtail-space-2025/speakers.html, we can probably do that by adjusting the urls.py and view. 

As for documenting the process, there was a lot of trial and error. So I think the best option is to build the archive we want and then I'll rework the scripts so we reproduce that hand crafted archive. That should leave us in a better position to archive the next one.

cnk:
I have tried 3 times to make the linter accept the README.md file. I am running prettier but apparently not with the versions required to make this pass. Some one else want to try? or shall I add that file to .prettierignore?

vossisboss:
> I have tried 3 times to make the linter accept the README.md file. I am running prettier but apparently not with the versions required to make this pass. Some one else want to try? or shall I add that file to .prettierignore?

I can give this a go tomorrow. Thank you for following up on this!

vossisboss:
@cnk Thibaud upgraded the stylelint package and the linter configurations. So I guess we need to make sure the config is set to 2.0.0 on our testing branches to pass the linting checks.

``` 
- stylelint@17.4.0
- '@wagtail/stylelint-config-wagtail@2.0.0' 
 ```

vossisboss:
@cnk All I found that was broken was the schedule links. It was a quick fix though. I'm going to push this to staging and test it there now before deploying.

vossisboss:
I had to shift around some static files to make collectstatic happy but I think I finally have everything where it needs to be. I'm going to go ahead and merge this. Thanks so much for contributing this @cnk !

PR #614, opened 2026-04-22
Upgrading Wagtail to 7.4rc

vossisboss:
We're installing Wagtail 7.4rc for testing purposes and to see if there are any broader issues that need to be resolved before the actual release.

### AI usage

No AI to be found here. All human-made code.


vossisboss:
We found a couple issues worth noting but are still working on seeing if they are reproducible in another demo before filing issues. The issues we found so far include:

1. The live preview stops working after the first few autosave revisions
2. The comparison option in the History tab leads to a `NoneType` error when comparing autosaved revisions (older revisions are fine).
3. The check for an absent Meta description is failing to clear after correcting the issue and saving the draft. Here are some screenshots of that issue:

Screenshot 2026-04-22 at 4 19 50 PM

Screenshot 2026-04-22 at 4 20 06 PM


We'll see if we can officially reproduce any of these and then log official issues.

vossisboss:
I couldn't reproduce any of these issues in the Wagtail nightly demo. So these appear to be issues specific to Wagtail.org. Or in the case of the preview, it might be an issue with my browser. There might be something to the meta description check issue in that it's possible there is a way our meta data is configured that was not anticipated by the code in core.

Stormheg:
Superseded by #617

PR #615, opened 2026-04-29
Add a suport runbook

nickmoreton:
Temporary Support Ticket https://torchbox.atlassian.net/issues?filter=-1&jql=assignee%20%3D%20currentUser()%20AND%20resolution%20%3D%20Unresolved%20order%20by%20updated%20DESC&selectedIssue=AHMJ-336 available only to Torchbox internal team

I will update the ticket reference once it becomes available.

### Description

As part of our support process at Torchbox we would like to add this document to the codebase to help support engineers with key information should we need to investigate a service outage or other services request.

### AI usage

Generated by Claude code from our in house support runbook template

### DIFF
--- a/docs/frontend/support-runbook.md
+++ b/docs/frontend/support-runbook.md
@@ -1,0 +1,348 @@
+# Support runbook
+
+This document describes possible ways in which the system may fail or malfunction, with instructions how to handle and recover from these scenarios.
+
+This runbook is supplementary to the [general 24/7 support runbook](https://intranet.torchbox.com/propositions/design-and-build-proposition/delivering-projects/dedicated-support-team/247-support-out-of-hours-runbook/), with details about project-specific actions which may be necessary for troubleshooting and restoring service.
+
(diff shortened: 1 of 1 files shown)


vossisboss:
Hi @nickmoreton. I don't have access to our Jira instance, so I don't have the full context of what this document would help with.

Did you generate this task to be specific to Wagtail.org? I'd just like a bit more information about why the task needs to live here in our repo vs living with the support team.

nickmoreton:
Hi @vossisboss Sorry only just seen this. I will get back to you on Monday

docs/frontend/support-runbook.md
```diff
@@ -0,0 +1,348 @@
+# Support runbook
```
vossisboss:
```suggestion
# Torchbox Support runbook
```

vossisboss (changes requested):
@nickmoreton This all looks good. I just have one tiny change to suggest so that people understand this file is for the Torchbox support time.

PR #624, opened 2026-08-21
Draft: Setup v3 api for blog page on Wagtail 8.0rc1

JakubMastalerz:
When applied this PR will enable the use of the new v3 write api with the `BlogPage`. 

It will also make `MarkdownField` contents sanitised to protect against malicious use of the API.

Note: Should probably be rebased onto Wagtail 8 release/upgrade


wagtailio/blog/models.py
```diff
@@ -148,6 +149,25 @@ class BlogPageAuthor(Orderable):
...
     ]
 
+    api_fields = [
+        APIField("author", writable=True),
+        APIField("name"),
+        APIField("job_title"),
+        APIField("url"),
+    ]
+
+    @property
```
vossisboss, line 159: Can you tell me why you're choosing to add properties here? I don't think they are strictly necessary for the API to function, so I'm curious what your thinking is.

vossisboss (changes requested):
@JakubMastalerz I know this is just in draft state but I was about to add these settings myself and was pleasantly surprised to find someone already did it.

I had one question for you. Also, now that #625 is merged, you can rebase this as well.

thibaudcolas:
@vossisboss I’ll close this as we’re not planning to continue the PR as-is, it was part of a trial of the Wagtail 8.0 release candidate

